PR #7, opened 2021-07-20
Ensure that the rhs column has a set width

timgws:
The default for `col` is to have max-width set to 100%, sadly, this means that if there is content (such as a large image) that takes more then `calc(100%-25%)` of viewport, then the menu on the RHS will not show.

This tiny PR sets a width for the column, which ensures that the menu will appear on the RHS. Ideally, there would be a check here to ensure that the `col-md-7` is only applied when the sidebar has content.

### DIFF
--- a/index.php
+++ b/index.php
@@ -99,7 +99,7 @@
             <?php endif; ?>
             <!-- Post List -->
             <div class="row">
-                <div class="col">
+                <div class="col col-md-7 col-xxl-9">
                     <!-- Grid Layout -->
                     <?php if (have_posts() ) : ?>
                     <?php while (have_posts() ) : the_post(); ?>


craftwerkberlin:
Thanks for this PR. `col` class is used for dynamic width of content if sidebar has no content. This is fine if you use just excerpt in post list. But you're right, if you using content view in post list, then sidebar moves to bottom when there are large images.

Making still the column dynamic with `col-md-*` classes could be done with JavaScript. But I would like to reduce that rather than expand it. So, if there is no sidebar, user can remove `col-md-7 col-xxl-9` classes. This can easily be done in a upgrade-safe way by copy index.php to child-theme.

PR #10, opened 2021-08-05
Build "current_page_parent" and "current-post-ancestor" check into the navwalker

justinkruit:
Based on the issue that @craftwerkberlin opened in the navwalker, https://github.com/AlexWebLab/bootstrap-5-wordpress-navbar-walker/issues/7.


craftwerkberlin:
@justinkruit Thank you!

PR #20, opened 2021-10-06
Update Bootstrap 5.1.2

crftwrk:


justinkruit:
As this merge will also update the version in style.txt, I'll wait a small moment. I want to add one extra check to the compiler, which is to add the bootstrap.scss file of Bootstrap source to the timestamp check. This would ensure that the scss is always compiled when only the Bootstrap files are updated.

PR #31, opened 2021-10-29
Built a "filter" for terms.php

Axos11:
I wrapped the "getCorrectedTerms" around all hooks in that file (maybe no necessary), but it should only affect "correct" formated parts. (Tested with sendinblue newsletter plugin)

### DIFF
--- a/woocommerce/checkout/terms.php
+++ b/woocommerce/checkout/terms.php
@@ -9,10 +9,59 @@
 
 defined('ABSPATH') || exit;
 
-if (apply_filters('woocommerce_checkout_show_terms', true) && function_exists('wc_terms_and_conditions_checkbox_enabled')) {
-  do_action('woocommerce_checkout_before_terms_and_conditions');
-
-?>
+if (apply_filters('woocommerce_checkout_show_terms', true) && function_exists('wc_terms_and_conditions_checkbox_enabled'))
+{
+  function getCorrectedTerms($html) {
+    $doc = new DOMDocument();
+    if ( !empty($html) && $doc->loadHtml($html))
+    {
+      $documentElement = $doc->documentElement; // Won't find the right child-notes without that line. ads html and body tag as a wrapper
+      $somethingWasCorrected = false;
+      foreach ($documentElement->childNodes[0]->childNodes as $mainNode) {
+        if($mainNode->childNodes->length && strpos($mainNode->getAttribute("class"), "form-row") !== false)
+        {
+          if(strpos($mainNode->getAttribute("class"), "required") !== false)
+          {
+            $mainNode->setAttribute("class", "form-row validate-required"); // You could try to keep the original class and only add the string, but I think that could ruin the design
+          }
+          else
+          {
+            $mainNode->setAttribute("class", "form-row woocommerce-validated");
+          }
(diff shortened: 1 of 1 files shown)


crftwrk (approved):
👍

justinkruit:
May I ask what exactly the point is of this code? 

Also, I recommend to make the functions in the Woocommerce functions file, not in a template file.

crftwrk:
This fixes hooked broken checkboxes in checkout. For example a sendinblue newsletter check.

I agree that it's better to solve that as a function instead of changing the WC template. Any idea?

crftwrk:
Thought about it again. @justinkruit You're right, this file will too hard to maintain if WC updates the terms.php and we should keep template files as simple as possible. Was too fast pressing the merge button, my bad. I reverted this PR.

@Axos11 Are you able to remodel this filter using hooks and add it to woocommerce-functions.php?

Axos11:
As far as i know, there are no filters for the hooks in terms.php. (if there are, just tell me and i will happily use them)
The issue came up with the modification of the terms.php file in the first place. So reverting the changes but still changing the "terms.php" file to a bootstrap syntax is not really an option in my oppinion.

You can move the 2 functions to a function.php file, thats not an issue at all. The only problem that still exists is, that there is not really any access to this content without modifying the terms.php file. (we could add a new filter for the used hooks there, but still. the terms.php file has to be edited and its basically the same :) )

crftwrk:
Of course, this filter is absolutely neccessary. This is just some brainstorming how to add it more smooth. I'm a bit worry about maintaining this file. 

However, terms.php is edited already. So, more editing is ok. What's about your idea to add a new filter to file and do function in woocommerce-functions.php? A filter there is just one line of code in terms.php.

Axos11:
If you just move the 2 functions "getCorrectedTerms()" and "captureHookOutput()" to the woocommerce-functions.php then you only have 3 changed lines of code which are :

- do_action('woocommerce_checkout_before_terms_and_conditions')
+ echo getCorrectedTerms(captureHookOutput('woocommerce_checkout_before_terms_and_conditions'));

- do_action('woocommerce_checkout_terms_and_conditions')
+ echo getCorrectedTerms(captureHookOutput('woocommerce_checkout_terms_and_conditions'));

- do_action('woocommerce_checkout_after_terms_and_conditions')
+ echo getCorrectedTerms(captureHookOutput('woocommerce_checkout_after_terms_and_conditions'));

The adaption to filters would be just a cosmetic thing. Actually the functionality would be just the same (at least i think so :) )

Ps: (I have to say, that i don't like the woocommerce approach there anyways. because they mix hooks with html code where only hooks would be as good :) )

crftwrk:
Yes, think this is better. Keeping terms.php slim and put function to woocommerce-functions.php will be much more cleaner. Want to create a new PR?

> (I have to say, that i don't like the woocommerce approach there anyways. because they mix hooks with html code where only hooks would be as good :) )

I disagree with many things in WooCommerce ;-)

justinkruit:
I think this would be the right approach, having the functions in woocommerce-functions.php. That way the terms.php is the most authentic, and should make updating it easier.

PR #33, opened 2021-11-08
Style.css to scss

crftwrk:
- Removed CSS fallbacks for old headers
- Splitted style.css content to smaller scss files and `@import` in `_bscore_style.scss`  
- Converted css to scss
- Tested and works fine


justinkruit:
Moving the scss files into a subfolder makes it so the current compiler setup won't check if those have been updated and recompile if necessary.

Do you want me to update the compiler, and also check for all the files inside the bootscore folder?

crftwrk:
That's true. But I think it's a pretty good way to cut them into logical pieces. Idea behind this that we can create a _woocommerce.scss collector file as well and catch this in the bootcommerce child. Then we have only two css file requests. In any later major version we can move compiled output from bootstrap.min.css to style.css. Then whole css is packed in one single file. 

What do you think? If you mean that's an good idea, do you want to update compiler?

justinkruit:
I also think it's a good idea, it'll keep our code cleaner. I'll update the compiler soon to make support for a bootscore folder.

Do we want to force that the bootScore scss is only in the main theme, and not in a child theme? Like we did with Bootstrap scss?

justinkruit (changes requested):
Compiler won't check if bootScore scss is updated to recompile. This needs to be addressed before being merged. I'll add to this PR in the upcoming days with that change.

crftwrk:
> Do we want to force that the bootScore scss is only in the main theme, and not in a child theme? Like we did with Bootstrap scss?

Yes, of course.

justinkruit:
> BTW you do not have to check all scss files in the bootscore folder. They contain just the content from style.css.

Actually, I would need to check them all. If only one is updated in a release, but nothing else, it won't force a recompile.

justinkruit:
We need to either have a version code in _bscore_style.scss to trigger recompiling (so the file is newer on release), or we need to add the bootscore map to the check. Both are valid options. What would you prefer?

crftwrk:
Do you mean adding a source map or a version in _bscore_style.scss? The version must be updated by hand if I‘m right. What do you prefer?

justinkruit:
I mean adding a version code in the file itself, which we need to update with each release. 

I'm personally more a fan of also checking all the files in the bootscore folder, but I also know that will be more file checks that happen each page load.

I like how you ask my preference, without you telling me yours when I asked 😂

crftwrk:
> I like how you ask my preference, without you telling me yours when I asked 😂

Yes, because I'm not sure what is the best way ;-)

However, the simplest way with less maintaining would always be my favourite.

1. We can merge all files back to _bscore_style.scss.
2. Or we can leave it like it is. Compiler starts anyway if one of the other files is changed. The core style of the theme should never be edited by a user. This is more for us to keep files organised and we know that we have to do something in another file if we changed something in theme's scss.  Content of this files doesn't change often.

What do you think?

Axos11:
Hello, I hope i don't disturb you here ;)
I read the part of the compiler and if I understand correctly, the compiler checks for changes on every pageload?
If that is the case and we don't want the users to edit ANY core files I think I would prefer a versioning of the bootstrap-main.css as "done" as possible to reduce any unnecessary load. Maybe we could also use a "flag" in that versioning for developing which forces the compiler to always check for changes.

If i got something wrong just tell me :D

crftwrk:
Hi @Axos11,

> I read the part of the compiler and if I understand correctly, the compiler checks for changes on every pageload?

Of course not ;-) Compiler starts if one of the .scss files in folder scss is changed, but not if file is stored in a subfolder. That means if you do changes in the bootscore subfolder, nothing happen, you must change something in for example _bscore_style.scss to view your changes. This can be a version number or anything else like @justinkruit suggested above.

Autocompiling this files is not so important in my opinion, because users edit their scss files in child and these files compiles automaticly if you do changes in your child.

I'm absolutely not sure what's the best way. We can merge all files to _bscore_style.scss than there will be no issues, but if we split them into smaller files and store them in subfolder, we're a bit more organised.

To be honest, no idea what's the best.

Axos11:
> Of course not ;-) Compiler starts if one of the .scss files in folder scss is changed, but not if file is stored in a subfolder. That means if you do changes in the /scss/bootscore/ subfolder, nothing happen, you must change something in for example _bscore_style.scss to view your changes. This can be a version number or anything else like @justinkruit suggested above.

thats what i actually meant. :) The check for changes is happening on every pageload, but the compiler only starts if changes are actually detected. My suggestion would be, that the timestamp of the creation of the compiled css file is compared to the timestamp of the css file. If they defer to much (more than 1s or so) the compiler runs.

I would still suggest a "hack" for developing to force recreation every time ;) for example a .dev file in the same folder which excluded from git. We could also make a .dev-sample which is included in git in which we explain the functionality of it.

As i think of the .dev file we could actually just remove the check of the bootstrap-main scss files completely with the .dev approach. (maybe we could use the wp-debug flag. but I'm not totally sure if thats a good idea)

> I'm absolutely not sure what's the best way. We can merge all files to _bscore_style.scss than there will be no issues, but if we split them into smaller files and store them in subfolder, we're a bit more organised.
I strongly prefer the splitting. It is really hard to maintain everything without it. 

@justinkruit One question about the compiler, does it do autoprefixing the css?

justinkruit:
> thats what i actually meant. :) The check for changes is happening on every pageload, but the compiler only starts if changes are actually detected. My suggestion would be, that the timestamp of the creation of the compiled css file is compared to the timestamp of the css file. If they defer to much (more than 1s or so) the compiler runs.

That is actually what happens. The code adds the last modified of the files together, which includes all the files directly in the scss folder, and the bootstrap.min.scss file (so a recompile happens when bootstrap updates), and compares that last modified to a string stored in the `set_theme_mod`. If the combined modified is higher than the stored value, it will recompile and save the new last modified timestamp. I'm talking about versions in the scss file, as this will create a new timestamp if, for example, we only update something in the bootscore scss file, and people update the files, it wouldn't trigger a recompile if we don't check the modified timestamp of those files.

The main discussion we're having is if we should:
1. Keep the files in bootscore scss folder separate, and add all to the modified check
2. Keep the files in bootscore scss folder separate, and add a version to _bscore_style.scss to trigger a new modified check on update
3. Combine the bootscore scss again into one file, and don't have a separate folder.

Do you have an opinion on that, @Axos11 ?

> I would still suggest a "hack" for developing to force recreation every time ;) for example a .dev file in the same folder which excluded from git. We could also make a .dev-sample which is included in git in which we explain the functionality of it.
> 
> As i think of the .dev file we could actually just remove the check of the bootstrap-main scss files completely with the .dev approach. (maybe we could use the wp-debug flag. but I'm not totally sure if thats a good idea)

One thing we can use is [the environment setting](https://developer.wordpress.org/reference/functions/wp_get_environment_type/) in WordPress to always force a recompile. So a check if the environment is development. You then just need to define the environment in the wp-config and you're ready.

> @justinkruit One question about the compiler, does it do autoprefixing the css?

It does not, as it is not an option of scssphp. We can implement it, but it would require loading in yet another php library to make that work.

crftwrk:
May I ask you why so complicated? In my opinion it is not necessary that these files trigger the compiler, because they are the source of the theme exactly like Bootstrap source. The Bootstrap source files do not trigger the compiler either. On a fresh install or doing something in child, for example the `_bscore_custom.scss` get a new timestamp and all source compiles.

justinkruit:
I think 2 will be the most resource friendly approach, and adding an environment check in the compiler will help to test changes when we're working on the bootscore scss.

Axos11:
I would also go with the 2nd option. And the environment check is also a good idea.

I use autoprefixing in all my workflows because i hate to prefix things manually but I am using my gulp workflow anyways for several reasons (autoprefixing, browsersync, criticalcss, ..) so I am maybe not the best person to ask there.
(Just thought that it is kind of essential and i think many "not experienced" devs could forget about it)

crftwrk:
We do 2

justinkruit (approved):
I've reverted the change that checks the bootscore scss files, and added a version to _bscore_style.scss. 

@crftwrk If you think it's fine like this, I'd say it's ready to be merged. I'll make a new pull request for the WordPress environment.

PR #36, opened 2021-11-13
Added redirect if offcanvas login failed

### LINKED ISSUE #35: Alert notices are landing on offcanvas
There is a small bug in my-account (woocommerce) page. Some alert notices are landing on offcanvas. You can test it when you fill in for example the billing address fields and leave a required field blank. After save, the notices show up on offcanvas.

crftwrk:
Closes #35 

This PR removes the hook for alerts in user offcanvas and redirects to my-account page if login failed.

### Motivation

Before logged in via user offcanvas and entered a wrong password or username, offcanvas stayed open and alert was displayed inside offcanvas. This makes problems if a user edited data in dashboard. For example when changing the address, the alert for failed or success were shown in offcanvas too.

Now the hook for alerts is removed and user will be redirect to my-account page when login failed instead and alert is shown there. 

- Redirect to my-account page after login in offcanvas failed
- Redirect to my-account page after login in checkout failed
- Redirect to home page after logout
- Alerts in dashboard are shown in the top of the page when changing data there

### Test

Code is already live on [bootscore.me](https://bootscore.me). 

- Try to login with wrong user or/and password, you will be redirected to my-account page and get a feedback there what's wrong. 
- Login in user offcanvas with correct user and password. 
- Click in user offcanvas to Addresses and change something. Alert will be shown on top of the page if failed or success.
- Click Logout in user offcanvas and you will be redirected to home page.

### DIFF
--- a/woocommerce/woocommerce-functions.php
+++ b/woocommerce/woocommerce-functions.php
@@ -242,7 +242,7 @@ function bootscore_wc_capture_hook_output($hookName) {
 
 // Redirect to my-account if offcanvas login failed
 add_action( 'woocommerce_login_failed', 'bootscore_redirect_on_login_failed' , 10, 0 );
-function bootscore_redirect_on_login_failed() : void {
+function bootscore_redirect_on_login_failed() {
   // Logout user doesn't have session, we need this to display notices
   if ( ! WC()->session->has_session() ) {
     WC()->session->set_customer_session_cookie( true );
--- a/woocommerce/woocommerce-functions.php
+++ b/woocommerce/woocommerce-functions.php
@@ -184,6 +184,7 @@ function custom_loop_product_thumbnail() {
 // Add card-img-top class to product loop End
 
 
+// Correct hooked checkboxes in checkout
 /**
  * Get the corrected terms for Woocommerce.
  *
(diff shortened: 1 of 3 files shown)


woocommerce/woocommerce-functions.php
```diff
@@ -236,3 +237,26 @@ function bootscore_wc_capture_hook_output($hookName) {
   ob_end_clean();
   return $hookContent;
 }
+// Correct hooked checkboxes in checkout End
+
+
+// Redirect to my-account if offcanvas login failed
+add_action( 'woocommerce_login_failed', 'bootscore_redirect_on_login_failed' , 10, 0 );
+function bootscore_redirect_on_login_failed() : void {
```
justinkruit, line 245: Is there a reason for the `: void` added in this function? Have not seen the usage of that at all, to be honest.

justinkruit (commented):
Just need some clarification, otherwise seems good to me.

woocommerce/woocommerce-functions.php
crftwrk: No, haven't seen it. If it works without - delete it.

justinkruit (approved):
All seems fine 👍

PR #39, opened 2021-11-20
Force scss compile when environment is development

justinkruit:
When `wp_get_environment_type()` returns that it is in development mode, either from reading the `define` in the wp-config.php or from reading the `getenv`, the scss will always be compiled, no matter if changes have been made.

To set the environment to development, add the following line to your wp-config.php
`define('WP_ENVIRONMENT_TYPE', 'development');`

This PR has been made in response to comments from @Axos11 on #33


crftwrk (commented):
LGTM👍

Issue #5, opened 2021-05-19
Issue with 2FA plugin (WooCommerce - form-login.php)


cemmos:
I'm technically using the bootCommerce 5 Child Theme, but the child theme doesn't replace the form-login.php file, so I figured this would be the best place to post about this issue.

After trying to figure out what the issue was, I discovered that the form-login.php file is the issue with trying to log in with the [Two Factor Authentication](https://wordpress.org/plugins/two-factor-authentication/) plugin. The 2FA works using the default Twenty Twenty-One theme as well as the Storefront theme from WooCommerce (as well as other themes, but those are the ones I've tried personally).

**What's supposed to happen:**
When you try to log in after entering your username and password and pressing enter or clicking the login button, the 2FA plugin then inserts an input where you type in your 2FA code. It does this by using jQuery, I believe.

**What happens with this theme:**
When using bootScore-5/bootCommerce, the 2FA plug is unable to insert the input — there seems to be a page load after hitting the login button, and the error "_The one-time password (TFA code) you entered was incorrect_" pops up since you're unable to enter the 2FA code. When I remove the `bootscore-5/woocommerce/myaccount/form-login.php` file and try again, the 2FA process works as it should.

I imagine it wouldn't just be that plugin alone that would break, so fixing the issue would likely be better for overall compatibility with other 3rd party plugins.

craftwerkberlin:
Hello Cemmos,

thank you for reporting, I will check that.

1. Do you use the offcanvas login or a page for login?

2. Have a link to your site to check and understand that issue?

Basti

Cemmos:
Hey, Basti. I tried both the offcanvas and the regular page, the issue appears on both. I just now set up a test site here after seeing your reply: [wp.dev.obstance.com](https://wp.dev.obstance.com/)

2FA can be turned on directly from the My Account page. You can either register a new user or use this test user that I set up and then turn on 2FA, then log out and try logging back in:

Username: testuser
Password: m1%b4Hq5ce5t

craftwerkberlin:
Hey Cemmos,

I see… It seems there is a missing hook in form-login.php. Can you please rename or delete this file in theme that I can check that?

Thank you

Basti

Cemmos:
Thanks, Basti. I've now removed the file.

craftwerkberlin:
Hello cemmos,

The Two Factor Authentication plugin needs premium subscription to implement it in WooCommerce, so I cannot check it here on my site. 

[...]

PR #41, opened 2021-11-23
Woocommerce-style.css to scss part 1

crftwrk:
This PR adds a (still empty) `_bscore_woocommerce.scss` collector file and prepares to merge the `woocommerce-style.css` to scss like we did with style.css. This file is deactivated in parent `_bootstrap.min.scss` and can be activated by removing the comments. bootcommerce-child has activated this file by default Pull request [#5](https://github.com/bootscore/bootcommerce-child/pull/5).

File is loaded before `_bscore_custom.scss` and can easily be overwritten.

We should users give time to add this file by `@import` on their existing bootcommerce-child theme installation. In 5.2.1.0 we can move content from `woocommerce-style.css` to scss safely.

### DIFF
--- a/css/scss/_bscore_custom.scss
+++ b/css/scss/_bscore_custom.scss
@@ -1,7 +1,3 @@
-/*!
- * Custom styles
- */
-
 // Add your own code here!
 // Because this file is compiled after Bootstrap, you're also able to use Bootstrap mixins for better code.
 // Some examples can be found here: https://bootscore.me/documentation/bootstrap-css-sass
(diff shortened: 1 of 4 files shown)


Lo1176:
Maybe it is the same thing in thankyou.php line 35 and 37
and in payment.php line 46

crftwrk:
@Lo1176 this is the wrong Pull request? You‘re welcome to open new PR anytime if you find something.

Axos11 (approved):

crftwrk (commented):
LG

PR #43, opened 2021-11-24
btn btn-outline-primary

Lo1176:


Axos11:
I just wanted to leave the comment that the outline version fits the design but its quite easy to overlook. Should't it have something like button-warning or button-danger?

Lo1176:
Yes a *btn-outline-danger* should be great

Issue #51, opened 2021-12-01
Safari scroll fix causes weird behaviour


Axos11:
this few lines of code let the browser scroll up to the top on every pageload (if I open a product page from a category page and then hit the back button (category page > product page > category page) then the page is loaded with the correct position and the .js lines let the position jump up and then down again. It's also possible to replicate the issue with just using F5 on a scrolled category page in chrome.

`if (window.location.hash) scroll(0, 0);
  setTimeout(function () {
    scroll(0, 0);
  }, 1);`

As i don't really understand the issue the lines should fix I'm not sure how to fix them (I mean without removing the lines totally :))

crftwrk:
You mean the smooth-scroll script? Created this for one-pagers https://bootscore.me/theme/scrollspy-onepager/#four

Safari is a weird browser and many things are working there not or not as good like in all other Browsers. This script is only there, because Safari is the only modern browser that doesn't yet support `scroll-behavior` https://caniuse.com/?search=scroll-behavior. Hope we can remove this script soon.

crftwrk:
As mentioned in PR https://github.com/bootscore/bootscore/pull/52, this does not work correctly on my devices and it seems too complicated for me. I'm tired of this script too and would suggest this:

- We remove script and use the default `scroll-behavior: smooth;`
- Describe the snippet in docs, so users can add it to their child **or**
- we make a plugin of it **or**
- we do address script and css (`scroll-behavior: auto;`) only to Mac Safari and all iOS devices and place it at the bottom of theme.js. 

I suggest to do this in 5.2.1 as well.

crftwrk:
Script can be removed by iOS 15.4 (expected March 2022) https://caniuse.com/?search=scroll-behavior

Issue #37, opened 2021-11-19
bootscore_ie_alert() gets cached


Axos11:
I just had the issue on a client site that the function bootscore_ie_alert() output gets actually pagecached if the first user to open the site after a cache clear is a IE User. Also, the function is not triggered if the page is already cached. As caching is very important for the sites speed i would suggest to move that check to an js function or an ajax call to prevent it beeing cached.

crftwrk:
I'm more a friend of removing this function. Bootstrap were the first who dropped the support for IE but many others like WooCommerce followed quickly. Created this function when doing the step from v4 to v5 to make clear that there is no support for this browser anymore. But now we have 2021 and there are [polyfills](https://github.com/coliff/bootstrap-ie11) to make Bootstrap 5 compatible with IE 11.

However, many users like this function. So, not sure to keep it or remove it. Addressed this point already to v6. A js version will give us more control to remove it later again instead using a hook and function. On the other hand I try to remove js wherever it's possible. So, no idea what's the best at this point.

Axos11:
I think the function kind of "helps" pushing people to modern browsers. And I agree that eliminating js where possible is a good idea (maybe could add it as a plugin?)

We could insert it the same way it is right now or as an Overlay. However, it has to be inserted into the page with js in my opinion because I have no idea how to avoid the page cache otherwise.

justinkruit:
I think the js isn't even that heavy, just an user agent check and go from there. You can even keep the alert in html and only remove d-none class when the user agent matches.

Axos11:
I would just try to put the alert as far as possible to the bottom to prevent Google from catching it somehow as a important component when it's crawling. Could work as a toast then.

crftwrk:
In an earlier version was an pure HTML and CSS solution using `<!--[if IE]>...<![endif]-->` and `@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {...}`. But I was afraid that Google could crawl that. So, decided to use a PHP function instead.

Toasts are very cool, but is IE able to show them? https://www.browserling.com/browse/win/7/ie/11/https%3A%2F%2Fgetbootstrap.com%2Fdocs%2F5.1%2Fcomponents%2Ftoasts%2F

Have no real IE to test here.

Axos11:
It sadly doesn't completely. It seems that with bootstrap 5.0.3 it worked but 5.1 kinda broke IE support completely. So the pages start to look really ugly if opened with IE.

[...]

PR #52, opened 2021-12-02
Fix ie warning fix - ios scroll

Axos11:
Here are my to .js fixe for the issues #37 and #51 .

The warning screen is not styled but functional. (some styling still should be done I think)

The scroll fix should work right away.

### DIFF
--- a/js/theme.js
+++ b/js/theme.js
@@ -51,6 +51,8 @@ jQuery(function ($) {
 2. Theme
 --------------------------------------------------------------*/
 
+if(checkForIE()){renderIeBrowserCheck()};
+
 jQuery(function ($) {
   // Smooth Scroll. Will be removed when Safari supports scroll-behaviour: smooth (Bootstrap 5).
   $(function () {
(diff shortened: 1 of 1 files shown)


crftwrk:
Hmm... we should split this in two different PR's to keep this separated. Have some questions:

## IE Alert

- There are no changes in hook in `header.php` and function in `template-tags.php`. Function is still there and can be cached. We should decide what we do. When we remove hook/function and switch to js version, we have to inform users that they have to remove hook in child as well.
- When removing function, we can add translation strings to `/languages/plugin-translations.php`
- Instead of inline-css we can use Bootstrap classes `position-fixed top-0 end-0 bottom-0 start-0`
- Agree with the approach

## Safari smooth-scroll

Doesn't work fine on my computer and phone and messed up theme.js in my opinion. Think we should decide what we do:

### Removing script

By removing this script we will kick out all Mac Safari and all iOS browers by default. But I'm open to do this and add this as a additional script to the docs. So everyone can decide what he do. In this case we have to remove

```scss
// scroll-behaviour: smooth not supportet by Safari. 
// jQuery smooth scroll script will be removed when Safari supports it
html {
  scroll-behavior: auto !important;
}
```
 in `css/scss/bootscore/_temporary.scss` and add offset for the navbar.

```scss
// Height of the navbar
html {
  scroll-padding-top: 55px;
}
```

### Keep this script until Apple supports `scroll-behavior`

The whole script is just a workaround. It's not optimal but does a great job in the past. I can live with the little bugs, because I know that the script will be removed anyway.

Bootstrap 5.2 comes closer, I would address both issues to 5.2.1. What do you think?

Axos11:
### IE fix

Okay, the ie popup was more like a suggestion to work on :). I would appreciate some styling input too :D

I never had to translate something in WordPress an js. What is the "best-practice" there? Do you have an example where something like that is already implemented in bootscore?
Todos:
- split PR
- I will remove the header hook with the next push. Actually I think we can tell everyone to remove it but as it should only be shown on ie anyways it is not breaking anything but just fixing it for those who actually update their child.
- I will adapt the classes instead of too much inline css :)

### scroll fix
What part is actually not working on your pc? Is the device not detected correctly or is something else happening?

crftwrk:
Closed by https://github.com/bootscore/bootscore/pull/53

PR #68, opened 2022-01-09
Add redirect to my-account after registration

pati1892:
After registration is completed or in case of an error, the My Account page should be displayed to show the user the error or the activation email note.


crftwrk (commented):
That definitely makes sense. Thank you.

PR #73, opened 2022-01-22
Fix accessibility issues

crftwrk:
It's not possible to add a `<label>` to searchform.php because form has no id. The searchform can't have an id because it will be displayed twice if it is in the top-nav-search widget position, mobile and large. In this case the id will be double. So, still remaining two accessibility errors.


crftwrk (commented):
All good

Issue #49, opened 2021-11-30
WooCommerce sub categories not displayed correctly


jonwebprocoder:
I am having problems with the formatting of subcategories on category pages. In the category i am setting this field

<img width="769" alt="image" src="https://user-images.githubusercontent.com/31352149/144032910-5645651d-3414-4272-96a2-a149611c4cc1.png">

But the formatting comes out all wrong

https://jonathanh113.sg-host.com/product-category/back-neck/

<img width="740" alt="image" src="https://user-images.githubusercontent.com/31352149/144032982-c60a7d23-d894-4f39-aba7-d09e6da5fd9d.png">

crftwrk:
See, there is a missing template. We'll fix this by the next update, but you can hotfix this easily.

### Hotfix

1. Create file `content-product-cat.php` and insert following code:

```php
<?php
/**
 * The template for displaying product category thumbnails within loops
 *
 * This template can be overridden by copying it to yourtheme/woocommerce/content-product-cat.php.
 *
 * HOWEVER, on occasion WooCommerce will need to update template files and you
 * (the theme developer) will need to copy the new files to your theme to
 * maintain compatibility. We try to do this as little as possible, but it does
 * happen. When this occurs the version of the template file will be bumped and
 * the readme will list any important changes.
 *
 * @see     https://docs.woocommerce.com/document/template-structure/
 * @package WooCommerce\Templates
 * @version 4.7.0
 */

if ( ! defined( 'ABSPATH' ) ) {
	exit;
}
?>
<div <?php wc_product_cat_class( 'col-md-6 col-lg-4 col-xxl-3 mb-4', $category ); ?>>
  
  <div class="card">
    
	<?php
	/**
	 * The woocommerce_before_subcategory hook.
	 *
	 * @hooked woocommerce_template_loop_category_link_open - 10
	 */
	do_action( 'woocommerce_before_subcategory', $category );

	/**
	 * The woocommerce_before_subcategory_title hook.
	 *
	 * @hooked woocommerce_subcategory_thumbnail - 10
	 */
	do_action( 'woocommerce_before_subcategory_title', $category );

    ?>
      <div class="card-body">
        <?php
    
        /**
         * The woocommerce_shop_loop_subcategory_title hook.
         *
         * @hooked woocommerce_template_loop_category_title - 10
         */
        do_action( 'woocommerce_shop_loop_subcategory_title', $category );

        /**
         * The woocommerce_after_subcategory_title hook.
         */
[...]

PR #84, opened 2022-02-01
Fix WooCommerce sub categories in loop

crftwrk:
- Fixes https://github.com/bootscore/bootscore/issues/49
- Button and badge done by filter (woocommerce-functions.php)
- `card-img-top` done by `@extend` in _wc_loop.scss


crftwrk (commented):
All good

PR #87, opened 2022-02-04
Bootstrap tabs to single product tabs

crftwrk:


crftwrk:
Tested with custom coded tabs and tabs plugin https://wordpress.org/plugins/yikes-inc-easy-custom-woocommerce-product-tabs/

Issue #97, opened 2022-02-11
SCSS make source map


justinkruit:
Generating a source map is good to debug stuff.

My idea is to have the source map being created and added to the css file, when the environment is set to Development. This to make sure your scss doesn't get that much published when you put the website live. What do you think of this, @crftwrk?

crftwrk:
Yes, that sounds good. Thank you.

About routing scssphp output to a different file name and folder. You want to do or should I do this?

PR #98, opened 2022-02-12
Add scss sourcemap support

justinkruit:
Connected to #97.

I've tested it on a local environment, and all seems to work, also for the child theme. @crftwrk could you try it out on a webserver environment? I currently don't have one set up for testing this.


crftwrk (approved):
Works like a charme. Very useful, thank you!

PR #105, opened 2022-02-19
Fix for scss .map when using subdirectory

### LINKED ISSUE #103: Source map doesn't show source in some installations
If `define('WP_ENVIRONMENT_TYPE', 'development');` is active, `main.map` is generated and linked in `main.css`. But it doesn't show source in dev console in some installations. In some installations it does, in some not. I tested it on different servers and version before and after moved `bootstrap.min.css` to `main.css`. May I messed something up in `scss-compiler.php`? @justinkruit any idea?

justinkruit:
This should fix #103. @crftwrk could you try this out on the installations you stumbled upon this issue? I've tried it in a test environment, and it seems to work fine now.

### DIFF
--- a/inc/scss-compiler.php
+++ b/inc/scss-compiler.php
@@ -40,9 +40,9 @@ function bootscore_compile_scss() {
 
   if ($is_environment_dev) {
     $compiler->setSourceMapOptions([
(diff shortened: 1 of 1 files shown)


crftwrk (approved):
Works perfect. Thank you!

PR #141, opened 2022-04-21
$labelledby does not exist

justinkruit:
I was getting a PHP notice that $labelledby does not exist. Checked with the default template, and it seems it indeed does not exist.


crftwrk (approved):
👍

PR #145, opened 2022-04-28
Make footer stick to bottom

### LINKED ISSUE #143: content min-height
Is there any reason why on markup.scss there is a min-height of 600px for the content ID ?
And how to remove that on child theme ?

justinkruit:
This should fix #143 for the future.


justinkruit:
When this gets approved, I'll soon do another run of reformatting the project. I've noticed some files being a bit messy, so just wanted to clean that up.

crftwrk (approved):

PR #149, opened 2022-05-03
Improve admin bar

crftwrk:
- Added `_admin_bar.scss`
- Moved `.logged-in.admin-bar` from `_header.scss` to `_admin_bar.scss`
- Added `_footer.scss`
- Moved sticky footer from `_markups.scss` to `_footer.scss`

We can handle sticky footer if user is logged in in `_admin_bar.scss` later.

### DIFF
--- a/scss/_bscore_style.scss
+++ b/scss/_bscore_style.scss
@@ -7,6 +7,7 @@
 @import "bootscore/breadcrumb";
 @import "bootscore/colors";
 @import "bootscore/comments";
+@import "bootscore/footer";
 @import "bootscore/forms";
 @import "bootscore/header";
 @import "bootscore/markups";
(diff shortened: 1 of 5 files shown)


justinkruit:
> We can handle sticky footer if user is logged in in `_admin_bar.scss` later.

What do you mean by that? The footer should already be at the bottom of the user is logged in.

crftwrk:
> The footer should already be at the bottom of the user is logged in.

Sorry, wrong communication, my mistake. What I mean is, when you are logged in, the admin bar pushes the content down. In result the footer is some pixels below the bottom of the browser. But that's super neatpicky and not possible to solve  because `html` is above the `<body class="logged-in admin-bar"...`. 

So just ignore it. This PR is just a tiny improvement in scss files, nothing special.

justinkruit (approved):
Just some lines being moved around, fine to merge. 👍

Issue #140, opened 2022-04-20
Update Bootstrap 5.2.0


crftwrk:
- [x] Update Bootstrap 5.2.0-beta1 https://github.com/bootscore/bootscore/commit/7389d0785d674eafccc9b74fbc45de32ea11d051
- [x] Update variables docs
- [x] Check Scrollspy and docs
- [x] Update header.php docs
- [x] Update FA6 docs
- [x] Update Google Fonts docs
- [x] Update translation docs
- [x] Dark mode Accordion
- [x] Remove `/scss/bootscore/_offcanvas.scss` https://github.com/bootscore/bootscore/pull/159
- [x] Rename page-blank-without-container to page-blank.php https://github.com/bootscore/bootscore/pull/163
- [x] Store page and single templates in folder https://github.com/bootscore/bootscore/pull/163
- [x] Remove navbar-light classes in header.php https://github.com/bootscore/bootscore/commit/54985949a75219db1e7139e21bca867619ed07e0
- [x] Added responsive offcancas to _admin_bar.scss https://github.com/bootscore/bootscore/commit/627abe95e946c8ab4a63cb7b418c9a5eddebc621
- [x] Change badge `text-bg-*` classes https://github.com/bootscore/bootscore/commit/ca17c9d617497689cb92e79358f8e2886d01122e
- [x] Pagination border hotfix https://github.com/bootscore/bootscore/commit/571c3aa953582b901a362f992c528d2ede7dd0f9
- [x] Modal bS Cookie Settings plugin hotfix https://github.com/bootscore/bootscore/commit/150fa4c1bcbdc93f1a2f88e33bed1ce70f1f74cb
- [x] Sidebar to responsive offcanvas https://github.com/bootscore/bootscore/pull/171
- [x] Disable/rewrite smooth scroll docs
- [x] Rewrite IE alert docs
- [x] Update child-themes 
- [x] Update Bootstrap 5.2.0 stable https://github.com/bootscore/bootscore/commit/da15257ff2d69e854198a618aa6843477250275e
- [x] Update version numbers https://github.com/bootscore/bootscore/commit/de7f69ebd6ccad0fed7ead497580c232180a808d
- [x] Revisit Pagination border hotfix https://github.com/bootscore/bootscore/commit/571c3aa953582b901a362f992c528d2ede7dd0f9, new hotfix https://github.com/bootscore/bootscore/commit/16eb38e8f6452df919d5ef9d0c0339e408400405
[...]

PR #155, opened 2022-05-16
5.2.0.0-beta1

crftwrk:
- Closes https://github.com/bootscore/bootscore/issues/140
- Includes Offcanvas fix https://github.com/twbs/bootstrap/commit/bb4f9d74fb399986eeeed07da10eded0908200b1
- Full change log https://github.com/bootscore/bootscore/issues/140
- Tested and works fine https://projekte.crftwrk.de/bs-5.2.0-beta1/

@justinkruit If you agree, I'm happy when you click the merge button. There is a drafted release and I'm happy too if you want to release this tomorrow morning. I do the blog post now.


crftwrk:
> You think it's a good idea to already make a release with BS 5.2.0 Beta 1? Checking the Projects page (https://github.com/twbs/bootstrap/projects/45) there are already some more bugs, so maybe it's best to wait a few more betas or the full release.

Agree. So I have more time for the blog. You're correct, let's wait.

Issue #47, opened 2021-11-29
Roadmap for scss files


crftwrk:
Since we have source in theme, I'm racking my brain how to merge them all to one single css at the end without disturbing users too much. I'd like to share some thoughts about a clear roadmap.

### 5.1.3.1

- [x] Move content from `style.css` to scss files. Done by https://github.com/bootscore/bootscore/pull/33
- [x] Add a new empty `_bscore_woocommerce.scss` to collect WooCommerce styles. https://github.com/bootscore/bootscore/pull/41
- [x] Inform bootcommerce user that they should **add** `@import "bscore_woocommerce";` to their child https://bootscore.me/blog/bootstrap/bootscore-5-1-3-1/
- [x] Inform users that they should **remove** `@import "fontawesome";` in their child https://bootscore.me/blog/bootstrap/bootscore-5-1-3-1/

### 5.2.0.0

- [x] Because users added `@import "bscore_woocommerce";`, we are able to move content from `woocommerce-style.css` to scss, delete file and remove enqueue snippet in `woocommerce-functions.php`. One file request less. https://github.com/bootscore/bootscore/pull/82
- [x] Because users removed `@import "fontawesome";` in child, we can delete this file. We should move content from `fontawesome.min.css` to somewhere in the subfolder `scss/bootscore/`. We can delete file and remove enqueue snippet in functions.php. One file request less. By doing this, we have more control to handle icon-font or inline-svg. Discussed here https://github.com/bootscore/bootscore/issues/46. https://github.com/bootscore/bootscore/pull/160
- [x] Adding a source map https://github.com/bootscore/bootscore/pull/98.

### 5.2.X.X

This is the hardest step, users have to edit enqueue script in child.

- [x] ~~At best, we set compiled output to `style.css`. So there is just one minifed css file at the end. Best performance, but hard to handle in child. https://github.com/bootscore/bootscore/tree/compiler-to-style.css~~
- [x] Better to handle in child. We can rename `_bootstrap.min.scss` to `main.css` or whatever. Because file contains now all css, name `bootstrap.min.css` is not correct anymore and confusing. https://github.com/bootscore/bootscore/pull/99

In conclusion, first collect all css to scss, then move the final file to somewhere else. Similar breaking changes in the past have shown that users will accept them if this serves the purpose and is logical. Think we are addressing around 400 - 600 active maintainers here.

@justinkruit @Axos11 any better suggestions?

Issue #46, opened 2021-11-25
Update to Font Awesome 6, switch to Bootstrap Icons or inline-svg?


crftwrk:
Font Awesome 6 is already in Beta3. Classes for icons has nothing to do with current installed Font Awesome 5 anymore. 

This seems to be the right point to think about to switch to Bootstrap Icons instead update to FA6. Because classes changed and we must change them anyway.

- Bootstrap Icons are lightweight, more than the half zipped size compared to FA.
- Both fonts (FA free) has 1500 + Icons
- After more than 10 years of using FA, BI can bring a fresh, new look.
- We can create a FA5 plugin for fall back if user want to use this.

| Font Awesome  | Bootstrap Icons |
| ------------- | ------------- |
| bootscore-main.zip: **2,1 MB**  | bootscore-main.zip: **851 KB**  |
| <img src="https://lh3.googleusercontent.com/pw/AM-JKLWfWsOapdCzblpWe1yqAd87rt5wSw2LqeVH5MoqpDZ9VaOtUJng_ckZZ4b41hNkJ3uMw6NSqY137-lVpw_fB3pBLLxrG1UMscU2ux2WEybBoP-njw9y4cwHvM8T_nANft1D305zZeEq0MJrNgieTddcuQ=w828-h112-no" width="414">  | <img src="https://lh3.googleusercontent.com/pw/AM-JKLXdLaiB8r8b184hrfDB-iWAbUSl9bdiNg7b7vZQia7rWvKJuVMXQ3hOy2UG1OjlN0IDlCP9mLXAh4ii-HOQncMmmPGDpmPkeyXIJ_1FtM_kCY92KeKtc34Af1DVzM1BMLsZFxxwTs4OzprUvjDOSjb9fg=w828-h112-no?authuser=0" width="414"> |

### Icons are used in
- **header.php** (search, user, cart, nav-toggler)
- **footer.php** (to-top button)
- **functions.php** (Breadcrumb)
- **searchform.php** (search button)
- **inc/template-tags.php** (comment-link, comments-count)
- **css/scss/bootscore/_alerts.scss**
- **woocommerce/product-searchform.php** (search button)
- **woocommerce/cart/cart.php** (trash icon)
- **woocommerce/cart/mini-cart.php** (trash icon)

In my opinion, Font Awesome has more visual power.

Axos11:
[...]

PR #160, opened 2022-06-01
Font Awesome to scss

crftwrk:
This PR pushes `fontawesome.min.css` to `_fontawesome.scss`

**Advantages**
- All previous css files are now scss and compiled to a single css file.
- There is no folder `css` or `css/lib/` in the repo anymore. Folder `css` and files will be created by scssphp.
- Less file requests
- Tested and works fine

**Disadvantages**
- The fonts must link to the main-theme folder `../../bootscore-main/fonts/fontawesome/` (see `scss/bootscore/_fontawesome.scss` line `6347` for example) otherwise this does not work in child-themes. 
- When renaming the main-theme folder (maybe someone wants to use it as an own framework), the path to fonts must be changed as well. This is something we can write down in the docs.

Closes https://github.com/bootscore/bootscore/issues/47
Closes https://github.com/bootscore/bootscore/issues/46


crftwrk:
Agree, we leave it as it is.

PR #163, opened 2022-06-08
Move page/post templates to subfolder

crftwrk:
This PR moves the page and single templates to subfolders (possible since WP 3.4).

- Added `page-blank.php`. This file will replace the ugly template name `page-blank-without-container.php` later.
- The template settings on existing sites fall back to default `page.php`. This can simply solved by storing the template files in the child's root folder (Update guide).


crftwrk:
> Maybe an idea to do all the other files too?

Of course. Do you have an idea how to solve this?

PR #170, opened 2022-06-13
Bootstrap v5.2.0 beta1

crftwrk:
@justinkruit I do not want to release this, but we need Bootstrap 5.2.0-beta1 to go further. I want to test it on some small sites to be prepared for the 5.2.0 stable release.


justinkruit:
Sure thing! What we could do is a separate branch with the beta release, if that's something you're interested in.

Issue #172, opened 2022-06-13
Move archive-*.php, author-*.php and category-*.php templates out of the theme into a separate repo


crftwrk:
All files are very useful but „dead“ files until you are copy them to your child and start renaming. The idea behind this issue is to keep only archive.php in theme and move all other templates to a separate repo. Users can download them and cherrypick which one they want to use.

- It‘s not possible to select them in the backend
- archive.php works for category and author too
- If an user want  to have a separate category.php, just copy it to your child
- This will remove 15 files in the theme root folder. 
- It‘s maybe more clear that child-theme can be extended by this templates.

What do you think?

rolzav:
II think it's a great idea. Personally, I vote +

rolzav:
I think it's a great idea. Personally, I vote + for it

justinkruit:
I was kinda hesitant, but one of my biggest annoyance of bootScore is that the files feel kinda cluttered, and this would definitely clean it up. We just have to decide what is gonna be the "main" look of bootScore.

crftwrk:
In v4 version, there was a cookie banner and even the preloader part of the theme. They went out step by step into plugins and since v5 I try to clean up the repo step-by-step again by reorganising the folders, files, renaming etc. without disturbing users too much. So, in my opinion the theme itself should be "bare" and clean but heavy extendable by additional files and plugins.

| Before  | After |
| ------------- | ------------- |
| ![before](https://user-images.githubusercontent.com/51531217/173632454-5571efda-8084-4cb7-9bb0-5d37f3b377a8.png)  | ![after](https://user-images.githubusercontent.com/51531217/173632529-b722d373-3dcd-491f-95df-70d5dfa0ad79.png)  |

justinkruit:
A thing I have in my own theme, is everything separated in template-parts in subfolders. So for example, for posts I gave /template-parts/post/post-single.php. Maybe something like that using subfolders would be nice? It would remove the need of folders like page-templates.

crftwrk:
I‘m absolutely in. But can they be overriden in child by storing a copy of the file there?

justinkruit:
I know you removed your question, but I checked anyway. As long as you use `get_template_part()`, it will check the child theme first.

crftwrk:
https://github.com/bootscore/bs-loop-templates

> As long as you use get_template_part(), it will check the child theme first.

[...]

PR #176, opened 2022-06-21
Remove unused loop templates

crftwrk:
This PR removes the unused loop templates from bootScore's root folder (looks much more cleaner now). Additional templates can be now downloaded here https://github.com/bootscore/bs-loop-templates

Closes https://github.com/bootscore/bootscore/issues/172

@justinkruit If you're ok with that, I'll be happy if you hit the merge button


justinkruit (approved):
👍

Issue #165, opened 2022-06-09
Release Asset Mismatch


bryceleue:
Hello,

BootScore is amazing! 

I am a GitHub noob, so there is a strong possibility I am confused, but there appears to be at least one noticeable difference between the bootscore-main.zip file and the Source code in the downloadable assets for the latest release.  

I was reviewing the changes made to js/theme.js as I anticipated they will affect my project, and the source code shows you removed a smooth scroll polyfill for Safari, but it is still present in the main file.  Is this supposed to be like this?

crftwrk:
Hello,

> BootScore is amazing!

thank you!

Yes, you got it, let me explain. There was a drafted release for v5.2.0.0 that was accidentally released by pressing the wrong button in February. This created the tag 5.2.0.0 and latest release 5.1.3.2 compares to this. 

It is not possible to remove tag 5.2.0.0 retrospectively. So, we screw up this point. The only file that matters in 5.1.3.2 release is the `bootscore-main.zip` and the only thing that changed compared to 5.1.3.1 is the offcanvas backdrop. 5.1.3.2 is also not visible here in source, because it's just the theme from bootscore.me with solved backdrop bug and uploaded manually.

Since 5.1.3.1 we did a lot of changes and improvements. For example the smooth scroll polyfill for Safari is not necessary since Safari/iOS 15.4 anymore, because they support now `scroll-behaviour: smooth;`.

You can download latest main branch here https://github.com/bootscore/bootscore/archive/refs/heads/main.zip, but note that you have to use latest child-themes as well. https://github.com/bootscore/bootscore-child/archive/refs/heads/main.zip and https://github.com/bootscore/bootcommerce-child/archive/refs/heads/main.zip. Docs changed as well but are not published yet.

If you want to preview the Bootstrap 5.2.0-beta1 in the theme, check here https://github.com/bootscore/bootscore/tree/update-bootstrap-520-beta1.

Does this help?

casualquick:
Yes, that clears it up.

Thank you!

crftwrk:
You‘re welcome 👍

crftwrk:
Sorry, messed up the release again. Don't care about this, `bootscore-main.zip` is the only important file. We will correct this in upcoming 5.2.0.0 release.

crftwrk:
This should be fixed by https://github.com/bootscore/bootscore/releases/tag/v5.2.1.0 and never happen again.

PR #189, opened 2022-07-19
5.2.0.0 release

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/140

This PR ships the 5.2.0.0 release. Tested and works fine. There is one issue with pagination component, all items have border-radius, hotfixed it here https://github.com/bootscore/bootscore/commit/16eb38e8f6452df919d5ef9d0c0339e408400405

@justinkruit If you agree, I'm happy when you just hit the merge button. There is a drafted release as well, but we have to choose a new tag according to https://github.com/bootscore/bootscore/issues/165. I'm tomorrow not on my computer but still have to write some update guides because too many things changed here. I suggest to release on Thursday July 21 and it will be a honor for me when you release it.


justinkruit:
Would it be alright if we set the release for tomorrow, July 22? I'd like to fix the pagination bug if possible before the release, but I didn't have the time yesterday.

crftwrk:
Ok, let's do it tomorrow, thank you.

justinkruit:
Alright, after testing it seems like this may be a scssphp problem, and not Bootstrap. Specifically here: https://github.com/twbs/bootstrap/blob/edf9c40956d19e6ab3f9151bfe0dfac6be06fa21/scss/_pagination.scss#L78
The `if` goes wrong somehow, and it falls back on the `else` which has a border radius on each item.

crftwrk:
Ok, we leave it as it is and release tomorrow? We can look later for a better solution.

PR #192, opened 2022-07-25
Make comments button to pluggable function

ucalegonte:
This change makes pluggable the function that displays the button for submitting comments. This way you can override it and customize in the child theme's functions.php file.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -427,11 +427,13 @@ endif;
 
 
 // Comment Button
-function bootscore_comment_form($args) {
-  $args['class_submit'] = 'btn btn-outline-primary'; // since WP 4.1    
-  return $args;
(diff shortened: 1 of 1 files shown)


crftwrk:
Thank you. But while we are here, I suggest to rename the function from `bootscore_comment_form` to `bootscore_comment_button`, because it's just the submit button and not a form.

```php
if (!function_exists('bootscore_comment_button')) :
  function bootscore_comment_button($args) {
    $args['class_submit'] = 'btn btn-outline-primary'; // since WP 4.1    
    return $args;
  }
  add_filter('comment_form_defaults', 'bootscore_comment_button');
endif;
```

Do you agree and want to change it?

PR #193, opened 2022-07-26
Pluggable comment button and pw form

ucalegonte:
I've renamed the bootscore_comment_form() function to bootscore_comment_button() for clarity.
I also made both the bootscore_comment_button() and the bootscore_pw_form() functions pluggable to allow them to be overridden in the functions.php file of the child theme.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -427,27 +427,29 @@ endif;
 
 
 // Comment Button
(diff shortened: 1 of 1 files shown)


crftwrk (approved):
Thank you!

PR #194, opened 2022-07-26
Offcanvas cart-items to list-group-flush

crftwrk:
This PR removes `border-bottom p-3` classes and uses `list-group list-group-flush` component for offcanvas cart-items instead.

There is also a small improvement with the height of `cart-list`. If there are many products in the cart, the scrollbars will not cover `cart-footer` anymore.

Preview https://dev.bootscore.me/shop/

@justinkruit If you agree, I'm happy when you just hit the merge button.


justinkruit:
That's a nice change! Good job

justinkruit (approved):

PR #199, opened 2022-07-31
Improve offcanvas cart-loader

crftwrk:
Just some tiny improvements. We can use Bootstrap classes instead, but to avoid breaking changes in header.php this is done by a bit CSS instead.


justinkruit (approved):
Looks good 👍

PR #201, opened 2022-08-08
Load style.css after main.css

crftwrk:
This is for users who are still afraid to touch a .scss file. So, they can do overrides in style.css without the `!important` rule.


justinkruit (approved):
Smart plan, should have done that before.

justinkruit:
I've also added a small commit myself, to clean up the `$modificated_bootscoreCss` `if` function a bit. You can quickly check that out before the merge.

crftwrk (commented):
Awesome! I'm happy when you just hit the merge button.

Issue #229, opened 2022-08-26
theme stylesheet is loading after main


ThePixelPixie:
Hi there,
I'm working in the main theme here (I'm using multi-site, and child themes will be used for the different sites therein, and need to be based on this main one).  In functions, I have moved the $modificated_styleCSS line below where both "main" and "fontawesome" are called, but it's still loading before main for some reason (yes, I've cleared cache and tried different browsers). When I go into inspector, "main" styles are over-riding "style". Not sure why. Any thoughts? This is my first time trying BootScore, and so far I'm very much enjoying it and learning a lot, despite having used Bootstrap for many years. So I'd really like to understand what's going on here.

Thanks in advance.

crftwrk:
Hello,

this is already fixed and merged https://github.com/bootscore/bootscore/pull/201

However, you should do your custom styles in `scss/bscore_custom.scss` instead in `style.css` to compile your styles to `main.css`. No SCSS skills or any other knowledge needed, simply add your styles to this file.

ThePixelPixie:
Ohhhhh...ok. Thank you. I didn't understand that about bscore_custom.scss being the one to edit. Also, I'd downloaded this theme just yesterday, so the change mentioned in #201, I'd assume, was already there. I will start working in custom.scss file instead.

crftwrk:
We added a notice to child‘s style.css https://github.com/bootscore/bootscore-child/blob/main/style.css but forgot to add it in main-theme. We will add this here too. Thank you for pointing this out.

You can just download latest main branch here https://github.com/bootscore/bootscore/archive/refs/heads/main.zip. So you can work with the fixed style.css loading order.

PR #232, opened 2022-08-29
Add scss notice to style.css

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/229


justinkruit:
Shouldn't we also remove the normalize.css part, as that isn't included in the theme?

crftwrk:
> Shouldn't we also remove the normalize.css part, as that isn't included in the theme?

Done

justinkruit (approved):

PR #233, opened 2022-08-29
Remove nav-link last-child padding

crftwrk:
This snippet added an edge to edge last nav-link to container if no header widget is active. I'm not sure if we should keep this or following more the Bootstrap way and delete it.


justinkruit:
I always just keep the padding. It helps with the click area as it's on the link, and when styling a background for it.

justinkruit (approved):

PR #236, opened 2022-09-04
Set checkout shipping table heading to top

crftwrk:
image190


justinkruit:
Yeah, that looks a little bit cleaner 👍

justinkruit (approved):

PR #247, opened 2022-09-08
Update Bootstrap 5.2.1

crftwrk:
Update is live on bootscore.me and works fine.

@justinkruit there is already a drafted release. Want to copy content and create new v5.2.1.0 release? There are two more PR's we should ship here, update scssphp 1.10.0 and a tiny WC checkout table improvement.


justinkruit:
Adding the release tag of Bootstrap for reference: https://github.com/twbs/bootstrap/releases/tag/v5.2.1

justinkruit:
@crftwrk Merged the other 2 PRs, will merge this and then make a new release file. 👍

justinkruit (approved):

Issue #248, opened 2022-09-08
EDIT "the_author_posts_link"


avdsm:
Good afternoon
I had to remove the update date of the post, and the image before the link to the author's profile.
I covered it with CSS.
Is there another way to be able to not only hide, but also change the icon in front of the author?
example here: https://vosmarket.am/recipes/havic-apur-manukneri/

PR #251, opened 2022-09-09
single.php meta author same as in loop

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/248

Discussion https://github.com/bootscore/bootscore/discussions/249

This PR changes  in `single-*.php` the `the_author_posts_link();` to `bootscore_author();` to use the same function there like in the loop.


justinkruit (approved):
Looks good 👍

PR #255, opened 2022-09-14
Update WooCommerce 6.9.0 templates

crftwrk:
WooCommerce changed in `form-login.php` template only the version number from `4.1.0` to `6.0.0`, nothing more has changed. You get an alert using an outdated template in backend.

This PR changes the version in the template file. We should release a quick v5.2.1.1 patch, that's why theme version number and readme.txt changed as well.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -7,7 +7,7 @@ Tags: featured-images, threaded-comments, translation-ready
 Requires at least: 4.5
 Tested up to: 6.0.2
 Requires PHP: 5.6
-Stable tag: 5.2.1.1
+Stable tag: 5.2.1.0
 License: MIT License
 License URI: https://github.com/bootscore/bootscore/blob/main/LICENSE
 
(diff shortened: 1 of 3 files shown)


justinkruit:
The version update is because they forgot it in November. The change they made hasn't been implemented in bootScore. It's just a string update, but nonetheless.

https://github.com/woocommerce/woocommerce/commit/b7c8e3ea68e5f74da36300b569ae8221d9896c47

crftwrk:
You mean to release it with Bootstrap 5.2.2? Agree.

justinkruit:
There are still changes in the template file that isn't in our version: https://github.com/woocommerce/woocommerce/commit/b7c8e3ea68e5f74da36300b569ae8221d9896c47

They updated the template file in november, but forgot to update the version string. This happened in the recent update, but we haven't included the changes they made back in November.

justinkruit:
When this is fixed, we can include this in a sub version release including the ie alert fix.

justinkruit (approved):

PR #256, opened 2022-09-14
Upstream main

crftwrk:


justinkruit:
Any reason you're doing this via PRs? As far as I know updating branches based on another branch doesn't require pull requests, especially if its updating from the main branch

crftwrk:
Yes, but it's just 2 clicks to update a dev branch to latest main changes via PR. So, why not staying up-to-date?

justinkruit:
Oh I totally agree to update the dev branch. I just mean, couldn't we just merge without the PR?

crftwrk:
GitHub creates a PR by itself. BTW it's just a rough sketch https://dev.bootscore.me/shop/ but let me know what you think.

PR #274, opened 2022-10-03
Feature breadcrumbs

sean-au:
Update standard and WC breadcrumbs to Bootstrap 5.2 spec.
Divider can now be styled with CSS :before

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -403,8 +403,9 @@ add_post_type_support('page', 'excerpt');
 if (!function_exists('the_breadcrumb')) :
 function the_breadcrumb()
 {
+
   if (!is_home()) {
-    echo '<nav aria-label="breadcrumb" class="mb-4 mt-2 py-2 px-3">';
+    echo '<nav aria-label="breadcrumb" class="breadcrumb-scroller p-3">';
     echo '<ol class="breadcrumb">';
     echo '<li class="breadcrumb-item"><a href="' . home_url() . '">' . '<i class="fa-solid fa-house"></i>' . '</a></li>';
     // display parent category names with links
--- a/functions.php
+++ b/functions.php
@@ -401,23 +401,30 @@ add_post_type_support('page', 'excerpt');
 
 // Breadcrumb
 if (!function_exists('the_breadcrumb')) :
-  function the_breadcrumb() {
-    if (!is_home()) {
-      echo '<nav class="breadcrumb mb-4 mt-2 bg-light py-2 px-3 small rounded">';
-      echo '<a href="' . home_url('/') . '">' . ('<i class="fa-solid fa-house"></i>') . '</a><span class="divider">&nbsp;/&nbsp;</span>';
-      if (is_category() || is_single()) {
-        the_category(' <span class="divider">&nbsp;/&nbsp;</span> ');
-        if (is_single()) {
-          echo ' <span class="divider">&nbsp;/&nbsp;</span> ';
-          the_title();
-        }
-      } elseif (is_page()) {
-        echo the_title();
+function the_breadcrumb()
+{
+  if (!is_home()) {
+    echo '<nav aria-label="breadcrumb" class="mb-4 mt-2 py-2 px-3">';
+    echo '<ol class="breadcrumb">';
+    echo '<li class="breadcrumb-item"><a href="' . home_url() . '">' . '<i class="fa-solid fa-house"></i>' . '</a></li>';
+    // display parent category names with links
+    if (is_category() || is_single()) {
+      $cat_IDs = wp_get_post_categories(get_the_ID());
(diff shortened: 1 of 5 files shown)


crftwrk:
Thank you for this PR. Yes, it's still the breadcrumb from v4 version, an update to v5 is absolutely a good thing. However, breadcrumb is wrapped in a horizontal scroller on mobile devices. I changed some classes in your function and added some css to keep this feature. Users can simply disable this by removing the classes from the wrapper.

Please switch to mobile view: 

- https://dev.bootscore.me/2009/07/02/edge-case-many-categories/
- https://dev.bootscore.me/product/beanie/

Let me know what you think.  

### Breadcrumb
```php
if (!function_exists('the_breadcrumb')) :
function the_breadcrumb()
{
  if (!is_home()) {
    echo '<nav aria-label="breadcrumb" class="breadcrumb-scroller mb-4 mt-2 py-2 px-3 bg-light rounded">';
    echo '<ol class="breadcrumb mb-0">';
    echo '<li class="breadcrumb-item"><a href="' . home_url() . '">' . '<i class="fa-solid fa-house"></i>' . '</a></li>';
    // display parent category names with links
    if (is_category() || is_single()) {
      $cat_IDs = wp_get_post_categories(get_the_ID());
      foreach ($cat_IDs as $cat_ID) {
        $cat = get_category($cat_ID);
        echo '<li class="breadcrumb-item"><a href="' . get_term_link($cat->term_id) . '">' . $cat->name . '</a></li>';
      }
    }
    // display current page name
    if (is_page() || is_single()) {
      echo '<li class="breadcrumb-item active" aria-current="page">' . get_the_title() . '</li>';
    }
    echo '</ol>';
    echo '</nav>';
    echo '</nav>';
  }
}
add_filter('breadcrumbs', 'breadcrumbs');
endif;
```

### WC Breadcrumb

```php
if (!function_exists('bs_woocommerce_breadcrumbs')) :
  add_filter('woocommerce_breadcrumb_defaults', 'bs_woocommerce_breadcrumbs');
  function bs_woocommerce_breadcrumbs() {
    return array(
      'delimiter'   => '',
      'wrap_before' => "<nav aria-label='breadcrumb' class='wc-breadcrumb breadcrumb-scroller mb-4 mt-2 py-2 px-3 bg-light rounded'>
      <ol class='breadcrumb mb-0'>",
      'wrap_after'  => '</ol>
      </nav>',
      'before'      => '<li class="breadcrumb-item">',
      'after'       => '</li>',
      'home'        => _x('Home', 'breadcrumb', 'woocommerce'),
    );
  }
endif;
```

### SCSS

https://github.com/bootscore/bootscore/blob/main/scss/bootscore/_breadcrumb.scss 

```scss
@include media-breakpoint-down(lg) {
  .breadcrumb-scroller {
    white-space: nowrap;
    overflow-x: scroll;
    overflow-y: hidden;
    -ms-overflow-style: none;
    overflow: -moz-scrollbars-none;
    scrollbar-width: none;

    &::-webkit-scrollbar {
      display: none;
    }

    .breadcrumb {
      flex-wrap: nowrap;
    }

    .breadcrumb-item {
      &.active {
        padding-right: $spacer;
      }

      + .breadcrumb-item::before {
        float: none;
      }
    }
  }
}
```

New file `scss/bootscore_woocommerce/_wc_breadcrumb.scss`

```scss
.wc-breadcrumb .breadcrumb-item:last-child {
  color: var(--bs-breadcrumb-item-active-color);
  padding-right: $spacer;
}
```

sean-au:
Looking great.

I would personally leave off "background-light rounded" as it has no effect
with the current background settings, however will be a bit of work to find
and fix if the background color is changed in a child theme.

crftwrk:
@sean-au cool, want to do quickly this changes?

sean-au:
Sure, but may not be til tomorrow as I'm off grid

sean-au:
I consolidated the margin and padding classes, used p-3 for the NAV and didn't override its child's mb-3 with m-0, has the same results but easier to target and override with CSS. 

I tested and I can't see that the wc-breadcrumb class is working, but maybe I am looking in the wrong place

crftwrk:
https://dev.bootscore.me/product/beanie-with-logo/

Default breadcrumb
Bildschirmfoto 2022-10-06 um 13 21 44

WC breadcrumb with `wc-breadcrumb` css
Bildschirmfoto 2022-10-06 um 13 22 04

WC breadcrumb without `wc-breadcrumb` css
Bildschirmfoto 2022-10-06 um 13 22 24

crftwrk (approved):
Merge this now to have no complications with your other PR's. We should keep the bg-light class because breadcrumb has the same style as before. I'll do it in a second PR.

PR #280, opened 2022-10-06
Feature-wc-store-notice

sean-au:


sean-au:
turns off default store notice, which is not bootstrap, or compadible with bootstrap
add a new store notice function, only if enabled and notice string not empty
changes the navbar to sticky rather than fixed
removes a div between header and nav, easier to target
adds a close button, and removes :before icon from store notice. 

NOTE: the child theme will override these changes - need to copy and paste woocommerce header into child theme to test

crftwrk:
Thank you for this PR. But we should not do this that way, because the store notice is a temporary thing and touching the `header.php` is sensitive. We can use the hook `<?php bs_after_primary(); ?>` for that and add some additional CSS for positioning. Is this an idea?

crftwrk:
We can use body class `.woocommerce-demo-store .fixed-top` to push content down like here https://github.com/bootscore/bootscore/blob/main/scss/bootscore/_admin_bar.scss.

sean-au:
Great I'll look into it I'm just remote for a few days but I'm will get
onto the PRs eventually

sean-au:
Hi @crftwrk, I have had a good look into this, here are the options (they all work but have limitations)

1. Original idea - inject with wp_body_open, change navbar to sticky-top, **notice has a close button** so can be removed if in the way. The sticky-top navbar is pushed down by the injected HTML, but goes back to the top of the page if the notice is closed.
2. Inject with bs_after_primary, or wp_footer (default wc behaviour)
  a) Display with fixed-top, inject an extra body class and push fixed elements and off canvas down **cannot have close button** 
  b) Display with fixed-bottom
    i) **has a close button**, gets in the way of the footer/body content
    ii) **has a close button**, add padding to footer equal to height of store notice. When store notice is closed, this padding is still present in the footer. **default WooCommerce behavior in store front theme**
    iii) **No close button**, gets in the way of the footer/body content, although the footer, and off canvas elements could be pushed up. Can have a close button, but not if your pushing items up with CSS


I think for UX considerations a close button is important in case the notice is in the way

crftwrk:
I completely agree that alert must have a close button. What's about this:

We can create a new hook inside the `fixed-top` above the `navbar`

```php
  <?php wp_body_open(); ?>

  <div id="page" class="site">

    <header id="masthead" class="site-header">

      <div class="fixed-top bg-light">

        <!-- Something like this -->
        <?php bs_top_nav(); ?>

        <nav id="nav-main" class="navbar navbar-expand-lg">

```

This is the same result as your first commit, but keeps the `fixed-top-class`. BTW this hook can be used for any other stuff as well. So, why not?

sean-au:
Sounds great I'll give it a go

crftwrk:
Maybe I have a new idea. We had a php function for warníng if user browsed site by Internet Explorer. Now we replaced this function by Javascript. But function hooked below the `</header>`

```php
  </header><!-- #masthead -->

  <?php bootscore_ie_alert(); ?>
```

https://github.com/bootscore/bootscore/blob/652bc3c411b154bf96361b32db689b99f9e9b295/header-woocommerce.php, please scroll down to the bottom. The result was this:



We can use this position for the shop notice. Then we have no trouble with navbar classes or overlapping content.

sean-au:
Great,Is the hook still there, and should I rename it? If yes, what should
I rename to?

crftwrk:
No, hook is removed. We have to create a new one, but it's simple and upgrade-safe.

#### header.php

```php
    </header><!-- #masthead -->
    
    <?php bs_after_masthead(); ?>
```

#### functions.php

```php
// Hook after #masthead
function bs_after_masthead() {
  do_action('bs_after_masthead');
}
```

#### Basic usage in child's functions.php

```php
function my_function() {
  echo '<div class="container mt-5"><p class="mt-4 alert alert-info">Hello Hook</p></div>';
}
add_action('bs_after_masthead', 'my_function', 5);
```

#### Result

screenshot

Now we can use this hook for store notice. What do you think?

sean-au:
Looks great, thanks for the help, I'll have a go at it tonight.

sean-au:
Can you have a look? I had it working, but then I am now having issues with my WP development environment.

crftwrk:
It is not working on my site. However, this PR has already conflicts with other (merged) PR's, there are 16 files changed. I suggest to close this PR to not mess up here anything and then start with two fresh PR's. One for the hook and the second one for the store notice. We will release 5.2.2.0 this weekend when Justin is ready to review. I would be happy if we can ship at least the hook by this release. Do you agree?

justinkruit:
@crftwrk should I wait with the release for this PR?

crftwrk:
@justinkruit let's release now and target this PR to next minor release. Bootstrap 5.2.2 has some very useful bugfixes.

PR #294, opened 2022-10-17
Fix nonstandard prefix for ::selection

### LINKED ISSUE #293: User selection is invisible!
Hello!

With bootScore 5.1.2,2 and WP 6.0.2 when a user selects something, the selection is invisible. This only seems to occur with bootScore, and I cannot reproduce on bootstrap itself.

Example: https://www.androidacy.com

androidacy-user:
Fixes #293

### DIFF
--- a/scss/bootscore/_colors.scss
+++ b/scss/bootscore/_colors.scss
@@ -15,7 +15,7 @@ Colors
 
 // pre tag
 pre {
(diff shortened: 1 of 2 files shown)


crftwrk (approved):
LGTM 👍

crftwrk (approved):
👍

PR #318, opened 2022-11-23
5.2.3.0

crftwrk:
Update to Bootstrap 5.2.3. Tested, works fine and is already live on bootscore.me. There is already a drafted release. @justinkruit if you agree, I'm happy when you just hit the merge button and publish release.


justinkruit:
Gonna try to make this release tomorrow

justinkruit (dismissed):

crftwrk:
We can also wait until 13th December. There is a WC update scheduled with some template changes https://developer.woocommerce.com/2022/11/23/woocommerce-7-2-beta-1/

justinkruit:
Oh that may actually be smarter, as BS 5.2.3 isn't _that_ special

crftwrk:
Updated WC 7.2.0 templates. Tested, works fine and is already live on bootscore.me. 

Quantity input is now visible on each product page, even product is out of stock or settings are limit to sell only 1 item. But amount in quantity input is not editable. Check here https://bootscore.me/shop/plugins/dark-mode/bs-dark-mode/. This is done by updated `quantity-input.php` https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/templates/global/quantity-input.php

@justinkruit if you agree, I'm happy when you hit the merge button and publish the release.

justinkruit (approved):
All looks good to me 👍

PR #329, opened 2022-12-19
Refactor offcanvas cart

crftwrk:
ajax-cart

This PR delivers a completely refactored ajax cart and has **no breaking changes since 5.2.0.0**.

Demo: https://dev.bootscore.me/shop/
Before & After clip: https://www.youtube.com/watch?v=IBsuoKNDrXk

Ajax cart is now compatible with all product types:

- Simple product
- Variable product
- Grouped product
- Affiliate product
- Out of stock
- Only 1 in stock
- Sold individually
- Add to cart buttons in other page content

### Add to cart buttons & Offcanvas cart

Offcanvas loader has been removed and buttons have now a loading-spinner instead. Offcanvas opens when product is added to cart.

### Alerts

Offcanvas shows now a success/danger alert when product is added/failed to the cart and removes them when offcanvas closes. 

#### Product page

Alerts are moved out from the top of the page into offcanvas cart.  These are the default WooCommerce alerts.

#### Loop

These alerts are not the WooCommerce alerts, because it seems that WC does not provide alerts in loop. Text can be translated like any other text in the theme.

There is no redirect to product page anymore if adding failed.

### Testing

Loop: https://dev.bootscore.me/shop/

#### Simple product

- Add “Album“ in loop to the cart
- Add “Album“ in product page to the cart https://dev.bootscore.me/product/album/

#### Variable product

- Click “Hoodie“ Select options in the loop
- Select variation and add to cart

#### Only 1 in stock / sold individually

- Add “Beanie“ in loop to the cart - success
- Add “Beanie“ one more time in loop - failed
- Remove “Beanie“ from cart and go to product page https://dev.bootscore.me/product/beanie/
- Add it twice to the cart. First success, second failed
- Remove “Beanie“ from cart

#### Grouped products

- Click “Logo Collection“ in loop
- Press + buttons to add one “Beanie“ and one or more  “Hoodie with Logo“, add them to the cart - success
- Add them again - Hoodie success, Beanie failed
- Click the Read more button to test the link if product is out of stock

#### Affiliate products

- Click “WordPress Pennant“ Buy on the WordPress swag store button in loop. You will be redirected to the WP swag store.
- Click button again on product page https://dev.bootscore.me/product/wordpress-pennant/. You will be redirected to the WP swag store.

#### Add to cart buttons in other page content

Click the “Album“ button below the `ul` https://dev.bootscore.me/2012/01/07/template-sticky/.

### Disable ajax cart

Because all scripts (except the mini cart in navbar) are now located in `ajax-add-to-cart.php`, it's very simply to disable ajax cart function. Disable "Enable AJAX add to cart buttons on archives" in backend WooCommerce > Settings > Products and add the empty function to your child's `functions.php`:

```php
// Disable Ajax Cart
function register_ajax_cart() {

}
add_action('after_setup_theme', 'register_ajax_cart');
```

### Known issues

Alert does not work in related products if product page is out of stock https://dev.bootscore.me/product/t-shirt/. This is done by this script which excludes the ajax cart if product is out of stock.

```js
$('a.ajax_add_to_cart:not(.outofstock a.ajax_add_to_cart)').on('click', function(e) {
````

Without these selectors, the “Read more“ loop button will not work.

If a product failed adding in the loop and then went directly to the cart/checkout page, the alert appears again at the top of the page. No solution for that yet. 

Closes https://github.com/bootscore/bootscore/issues/215
Closes https://github.com/bootscore/bootscore/issues/319
Closes https://github.com/bootscore/bootscore/issues/277
Fixes https://github.com/bootscore/bootscore/issues/200#issuecomment-1206965794
Fixes https://github.com/bootscore/bootscore/pull/205
Fixes https://github.com/bootscore/bootscore/pull/209
Fixes https://github.com/bootscore/bootscore/discussions/270
Fixes https://github.com/bootscore/bootscore/issues/154

Let me know what you think!

### CODE REVIEW

**Comment by justinkruit** [comment]
Oh god that's a lot to check 😂 
I'll soon take a full look at this!

**Comment by crftwrk** [comment]
Don't hurry up, it's not urgent 😎

**Comment by justinkruit** [comment]
Planning to take a good look at this this weekend.

Did you also check how well this works when Ajax is disabled? At work we use the bootScore woocommerce resources, and sometimes we need to disable ajax to make more advanced and custom systems.

**Comment by justinkruit** [comment]
Bug found: when going back a page, the cart isn't loaded with content.

**Comment by justinkruit** [comment]
Also, I see we're still using the default woocommerce alerts. Any chance we could kinda revamp that? The icon is set with the font version of Font Awesome, although I haven't used those for almost years now 😂 I've completely switched to the JS version, and needed to fall back on the woocommerce icons even though they don't look as good.

**Comment by crftwrk** [comment]
> Bug found: when going back a page, the cart isn't loaded with content.

Urgh, yes. Bug appears in actual ajax-cart https://bootscore.me/shop/ as well.

But it seems that this is a Woo mini-cart widget issue. Added the cart widget in sidebar here https://dev.bootscore.me/shop/. Always when you hit the go back button, cart is empty. You must add something new or reload the page. Bug appears even if ajax-cart is completely disabled. Think this has nothing to do with this PR.

> Also, I see we're still using the default woocommerce alerts. Any chance we could kinda revamp that?

FA icons to alerts are added here https://github.com/bootscore/bootscore/blob/main/scss/bootscore/_alerts.scss. I'm open to remove them in a second PR.

**Comment by crftwrk** [comment]
> Bug found: when going back a page, the cart isn't loaded with content.

We can use a script to force reload when back button is clicked. This one seems to work https://dev.bootscore.me/shop/:

```js
// Refresh page on browser back button 
// https://stackoverflow.com/questions/43043113/how-to-force-reloading-a-page-when-using-browser-back-button
window.addEventListener("pageshow", function (event) {
  var historyTraversal = event.persisted ||
    (typeof window.performance != "undefined" &&
      window.performance.navigation.type === 2);
  if (historyTraversal) {
    // Handle page restore.
    window.location.reload();
  }
});
```

justinkruit:
Isn't there a way to just refresh the cart, like a js event?

And for the fontawesome part, I know they're being set there, but that only works when you load the font. But the JS version replaces the i tag with an svg so those fonts aren't being loaded. Maybe we could load the svg directly there like bootstrap does.

crftwrk:
Figured out that this happens only on Chrome and Opera. Safari, Edge and Firefox works fine (on my machine). But this is already a known issue https://github.com/woocommerce/woocommerce/issues/32454. 

This workaround works fine https://github.com/woocommerce/woocommerce/issues/32454#issuecomment-1255015483:

```js
var isChromium = window.chrome;
if (isChromium) {
  $(window).on('pageshow', function (e) {
    if (e.originalEvent.persisted) {
      setTimeout(function () {
        $(document.body).trigger('wc_fragment_refresh');
      }, 100);
    }
  });
}
```


Font Awesome: We had this already here https://github.com/bootscore/bootscore/issues/46#issuecomment-979037213, but nothing happen?

crftwrk:
Fixed in https://github.com/bootscore/bootscore/pull/329/commits/e8bea45eb8c2c54694ae7fb4933118e4a16c92cb

justinkruit:
Did some more testing, and all looks good! Really impressed by it 😄 

I'll be doing some quick formatting tests and then it's ready to be merged.

crftwrk:
> I'll be doing some quick formatting tests and then it's ready to be merged.

Perfect ;-) But we should wait to merge until next week Wednesday, because on Monday a new Woo release with template changes (again) is scheduled. So, we can update templates on Tuesday and do a minor release 5.2.3.1 then. The cart we should ship with Bootstrap 5.3 if stable or beta, agree?

Mabye you can have a quick look to https://github.com/bootscore/bootscore/pull/332 and https://github.com/bootscore/bootscore/pull/344, which can be merged now. This are just minor improvements.

justinkruit:
While checking out the code, I noticed the js of the mini cart is getting quite long. Maybe it's a good idea to put it in a file instead of having it inline?

And I'll check out the others one first then, as most changes are in the js 👍

crftwrk:
Yes, it's long. But because JS and PHP is in one file it's very simple to deactivate with one hit. Translation strings will not work that way too if we put JS in an own file. And of course, ajax is direct on the server, no more file request.

crftwrk:
What do you think, should we ship the cart with the next Woo release or wait?

justinkruit:
In that case, would it maybe be cleaner to have the js in a separate file but echo the content directly with a file_get_contents()? That way the code looks a bit cleaner and you still have the direct js in the footer.

crftwrk:
https://dev.bootscore.me/shop/ seems good 👍

crftwrk:
If you agree, we merge now and release it next week.

crftwrk:
Want to press the merge button?

PR #332, opened 2022-12-20
Clean up single-*.php's and page-*.php's

crftwrk:
This just a small cleanup, nothing special. In conclusion, the comments moved to `.entry-footer` div.

In pages, I found this:

```php
<?php wp_link_pages(array(
  'before' => '<div class="page-links">' . esc_html__('Pages:', 'bootscore'),
  'after'  => '</div>',
));
?>
```

To be honest, no idea why this is there. Guess it's still from an very early v4. Just deleted it.

Changed `<?php the_title('<h1>', '</h1>'); ?>` into `<h1><?php the_title(); ?></h1>` for better understanding.

Removed comments. Think we do not need them ;-)

```
<!-- Content -->
<?php the_content(); ?>
```


justinkruit (approved):
All seems good to me 👍

PR #344, opened 2023-01-05
Add show-password to all pages in user offcanvas

crftwrk:
This PR adds the missing show-password in user offcanvas in non WooCommerce pages. Tested and works fine, already live on https://bootscore.me/.

## Before

| non Woo pages  | Woo pages |
| before  | shop-page  |

## After

| non Woo pages  | Woo pages |
| after  | shop-page  |


justinkruit (approved):
Great work as usual 😉

PR #350, opened 2023-01-08
Prevent self-closing offcanvas if dropdown has a nested dropdown (modified nav-walker with 3 or more levels depth)

kreincke:
https://github.com/bootscore/bootscore/discussions/347

### DIFF
--- a/inc/class-bootstrap-5-navwalker.php
+++ b/inc/class-bootstrap-5-navwalker.php
@@ -1,7 +1,8 @@
 <?php
 
 /*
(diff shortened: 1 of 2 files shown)


crftwrk:
Thank you! But please do not change the nav-walker, because 3 or more levels depth is not supportet by Bootstrap and we follow straight the Bootstrap's way. Changing the nav-walker is something what you can do in your child.

crftwrk (approved):

kreincke:
Thanks, got it. And I will keep in  mind your reasonable position

PR #365, opened 2023-01-18
Add spdx licensing tags

kreincke:
I've added the following text in the most php files I found. But ignored some deeper embedded files and some (s)css files ignored of which I could see that they have been created by the bootscore team.

### DIFF
--- a/404.php
+++ b/404.php
@@ -3,6 +3,9 @@
 /**
  * The template for displaying 404 pages (not found)
  *
+ * SPDX-FileCopyrightText: © 2021 Bastian Kreitrer (crftwrk) <pst@crftwrk.de>
+ * SPDX-License-Identifier: MIT
+ *
  * @link https://codex.wordpress.org/Creating_an_Error_404_Page
  *
  * @package Bootscore
--- a/archive.php
+++ b/archive.php
@@ -3,6 +3,9 @@
 /**
  * The template for displaying archive pages
  *
+ * SPDX-FileCopyrightText: © 2021 Bastian Kreitrer (crftwrk) <pst@crftwrk.de>
+ * SPDX-License-Identifier: MIT
+ *
  * @link https://developer.wordpress.org/themes/basics/template-hierarchy/
  *
  * @package Bootscore
--- a/comments.php
+++ b/comments.php
@@ -3,6 +3,9 @@
 /**
  * The template for displaying comments
  *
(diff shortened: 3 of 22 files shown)


crftwrk:
Thank you! Just have a few questions, because this new to me and I do not know how it works.

`* SPDX-FileCopyrightText: © 2021 Bastian Kreitrer (crftwrk) <pst@crftwrk.de>` 

Since I moved this project out of my personal profile into an orginazation profile, the license owner is not me anymore, the owners are we all who contribute to this project. Should it not be like this?

`* SPDX-FileCopyrightText: © 2021 bootScore (bootscore) <store@bootscore.me>`

First release was 2018, why there is 2021?

Files that should excluded:

- All files in folder `fontawesome` 
- `js/lib/bootstrap.bundle.min.js`
- All files in folder `inc/scssphp` 
- All files in folder `scss/bootstrap` 
- All files in folder `woocommerce` 

Correct?

kreincke:
Let me start with the simple questions:

a) I quickly review the github history and found 2021. If 2018 is a better year, you should take that. The best would be, if you can remember when you have created a file initially. If so, take that value.

b) Yes, any third party software which are taken over from other projects should be excluded. (You are not the copyright owner and you should not claim to be. ) I should have ignored the directories you mentioned - except class-bootstrap-5-navwalker.php which I assigned Alex WebLab. If that's not his real name, replace it.

c) Now the more important topic of copyright ownership:

Yes, you can do so. Normally, the first developer, who creates a file, is the file owner and copyright owner. In the beginning of free software - before github - each change was documented inside the files and commented by the copyright owner(s() of the change. This describing comment - if necessary - also contained devianting copyright statements. But nowadays, the change history is hold inside the git history. 

I did not really know who has created the files. Thus, I assumed you ;-) But if all of you agree to be covered by the label 'bootscore' everything is fine.

So, your solution is the better for giving your team the credit - as long as you prove who 'bootscore' in a sense of a group is. As far as I can see you automatically create a readme.txt containing the names of all contributors. Maybe you can add a simple remark inside of README.md that '© 2018 bootScore' to the contributors = copyright owners mentioned in readme.txt.

The function of the copyright statement is to indicate that the licensing statement is integrated by the copyright holder(s). Free software needs copyright owners and a licensing statement. (https://www.gnu.org/licenses/why-assign.html)

Yes, I know this is minor issue compared with the importance your work ass developer  has. But it is important for maintaining our biotope of free software.

with best regards KR

crftwrk:
Let me think about it for a while

Issue #333, opened 2022-12-20
Reduce WooCommerce templates


crftwrk:
Since WooCommerce 6.9.0 there are a lot of updated templates in every release, which is hard to maintain. Think we can reduce some templates using filter, `@extend` or add some CSS instead overriding them in theme. 

Potential candidates are `loop/add-to-cart.php`, `notices/error.php`, `notices/notice.php`, `notices/success.php` and a lot more because mostly it's just a button class or alert changed.

We can start this after https://github.com/bootscore/bootscore/pull/329 is merged.

Currently there are 59 templates used in theme. Here is a tracking list to review each file:

https://github.com/bootscore/bootscore/pull/372
- [x] ~~/woocommerce/notices/error.php~~ https://github.com/bootscore/bootscore/commit/a629ff777e651174f22b7192f4b7a6891e5251d0
- [x] ~~/woocommerce/notices/notice.php~~ https://github.com/bootscore/bootscore/commit/a629ff777e651174f22b7192f4b7a6891e5251d0
- [x] ~~/woocommerce/notices/success.php~~ https://github.com/bootscore/bootscore/commit/a629ff777e651174f22b7192f4b7a6891e5251d0
- [x] ~~/woocommerce/loop/no-products-found.php~~ https://github.com/bootscore/bootscore/commit/8a4f0ae25f1a320b07b250bb64de0de94f34d9d2
- [x] ~~/woocommerce/myaccount/downloads.php~~ https://github.com/bootscore/bootscore/commit/8a4f0ae25f1a320b07b250bb64de0de94f34d9d2
- [x] ~~/woocommerce/myaccount/form-add-payment-method.php~~ https://github.com/bootscore/bootscore/commit/8a4f0ae25f1a320b07b250bb64de0de94f34d9d2

https://github.com/bootscore/bootscore/pull/386
- [x] ~~/woocommerce/cart/proceed-to-checkout-button.php~~ https://github.com/bootscore/bootscore/commit/c9e0274e5f1e6f226977d760667d52a9870928f5
- [x] ~~/woocommerce/single-product/add-to-cart/external.php~~ https://github.com/bootscore/bootscore/commit/c9e0274e5f1e6f226977d760667d52a9870928f5
- [x] ~~/woocommerce/single-product/add-to-cart/grouped.php~~ https://github.com/bootscore/bootscore/commit/c9e0274e5f1e6f226977d760667d52a9870928f5
[...]

PR #372, opened 2023-01-20
Remove Woo notice templates

crftwrk:

This PR does it with all alerts in templates.

- Replaced all `alert alert-*` classes with default Woo `woocommerce-info`, `woocommerce-error` and `woocommerce-message`. 
- Using `@extend` to show them in Bootstrap's style
- 6 files that only had changes in the alert classes has been deleted
- Minor margin class improvement in checkout 

This has no impact on the frontend. Tested, works fine and is already live on bootscore.me. Just click through the store and check alert classes in cart, checkout and dashboard.

@justinkruit if you agree, I'm happy when you just hit the merge button. BTW, maybe I have an idea how to solve alert icons with Fontawesome's JS version, but let's do that later.


justinkruit:
I'll look into it this weekend!

Also, I said I was gonna look into the FA thing so I'll also promise to try some stuff out for that 😄

justinkruit (approved):
All looks good to me 👍

PR #375, opened 2023-01-28
Add more control over scss compiler

justinkruit:
This will add more control for the user over the SCSS compiler. You'll be able to disable the environment check (don't always compile in development or local) or disable the compiler completely.

**Priority of checking:**
1. If filter has been added
2. If a constant has been set
3. Fall back to `false`

Below the new filters, constants, and how to use them.

### Disabling the compiler
* Filter: `bootscore_scss_disable_compiler`
  * `add_filter('bootscore_scss_disable_compiler', '__return_true');`
* Constant: `BOOTSCORE_SCSS_DISABLE_COMPILER`
  * `define('BOOTSCORE_SCSS_DISABLE_COMPILER', true);`

### Disabling the environment check
* Filter: `bootscore_scss_skip_environment_check`
  * `add_filter('bootscore_scss_skip_environment_check', '__return_true');`
* Constant: `BOOTSCORE_SCSS_SKIP_ENVIRONMENT_CHECK`
  * `define('BOOTSCORE_SCSS_SKIP_ENVIRONMENT_CHECK', true);`

Fun additional thing:
Run the following command in the folder of the theme, and compile your styling without the need of the PHP Compiler. Note that it's formatted for Windows, and probably need to make small path changes for other platforms.
```
sass .\scss\main.scss .\css\main.css -w -I ..\bootscore-main\scss\ --no-source-map
```

### DIFF
--- a/inc/scss-compiler.php
+++ b/inc/scss-compiler.php
@@ -14,6 +14,12 @@ use ScssPhp\ScssPhp\Compiler;
  * Compiles the scss to a css file to be read by the browser.
  */
 function bootscore_compile_scss() {
+  $disable_compiler = apply_filters('bootscore_scss_disable_compiler', (defined('BOOTSCORE_SCSS_DISABLE_COMPILER') && BOOTSCORE_SCSS_DISABLE_COMPILER));
+
+  if ($disable_compiler) {
+    return;
+  }
+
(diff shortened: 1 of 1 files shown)


crftwrk (approved):
Great, thank you 👍

PR #377, opened 2023-01-30
Add an includable font awesome short code function

kreincke:
as discussed this morning: 

- inc now contains a file fa-shortcode.php containing the code [as literally been tested this afternoon]
- functions.php contains the respective include statement

Feel free to find a better integration

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -8,6 +8,7 @@
  * @package Bootscore
  */
 
+include 'inc/fa-shortcode.php';
 
 // WooCommerce
 //require get_template_directory() . '/woocommerce/woocommerce-functions.php';
--- a/inc/fa-shortcode.php
+++ b/inc/fa-shortcode.php
@@ -1,0 +1,14 @@
+<?php
+/*
+ * (c) 2023 Karsten Reincke
+ *  SPDX-License-Identifier: MIT
+ */
+function insert_bsfa($atts){
+  $atts = (array) $atts;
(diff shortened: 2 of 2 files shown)


crftwrk:
This is a good one 👍, but:

We do not need an additional file for this snippet. I understand, because it's about the license thing https://github.com/bootscore/bootscore/pull/365. But if we start to add a new file for every tiny change, new contributors will do the same and theme get messed up.

Better to add this in `functions.php` direct below the Fontawesome preload snipped in line `301` to keep all Fontawesome related scripts in one place.

Do you agree?

kreincke:
I totally agree! I hesitated to insert it into functions.php directly because I did not know whether you want to have in the main functions.php (wherein I would put it) or int the template for the child themes. 
Kindly regards kr
(BTW: So, in this case it was not a matter of the SPDX licensing headers ;-) But that has its advantages. By this statement, you know that I gave you the code under the terms of the MIT license. Informally, we knew that all the time. But now it is explicitly done. And provable ;-) )

crftwrk:
Want to change this?

kreincke:
Sorry - I did not see this ask. So, what's the best way for you? Shall I wait until 5.2.3.2 is out and reset the pull request to that base, or shall I do it now?

kreincke:
Yes, please do so. I will close all my clones too. And I will restart  the pr this weekend

PR #386, opened 2023-02-02
Remove Woo btn templates

crftwrk:

This PR does it with most `btn` in templates. There are still a few `btn` left, but I want to breakdown this into smaller pieces.

- Replaced `btn btn-*-*` classes with default Woo button classes
- Using `@extend` to show them in Bootstrap's style
- 8 files that only had changes in the button classes has been deleted
- Files that only contained a button class plus one utility like `w-100`, utility has been replaced by custom CSS

This has no impact on the frontend. Tested, works fine and is already live on bootscore.me. Just click through the store and check button classes in cart, checkout and dashboard.

BTW, WooCommerce has horrible selector classes, maybe SCSS looks a bit weird, but it's good.


justinkruit (approved):
Been a bit busy. All looks good!

Issue #397, opened 2023-02-12
Duplicated closing nav tag in the_breadcrumb()


patchampoux:
The line to remove is the 424th in functions.php.

crftwrk:
Urg, you‘re right. Want to create a pull request?

PR #398, opened 2023-02-12
Remove duplicated nav closing tag in the_breadcrumb().

patchampoux:
Deleted line 424 in fucntions.php.


justinkruit:
Will fix #397

PR #404, opened 2023-02-20
Make pagination translation-ready

patrickcognitif:
Registers strings in bootscore_pagination() in bootscore text domain.


crftwrk:
Thank you! Just two things:

- Please use 2 spaces for formatting instead tab.
- While we are here, we can remove `ft-wpbs` class, this isn't needed.

Tried to change both directly in your PR, but it seems that this is not working.

crftwrk (approved):
Looks good 👍

PR #410, opened 2023-02-28
Use WC strings instead of bootScore in mini cart

justinkruit:
Make it so some of the translations are being loaded from Woocommerce. Saves us making translations.


justinkruit:
Completely forgot I made this draft a month ago 😂

crftwrk (approved):
👍 Awesome

Issue #312, opened 2022-11-14
Search doesn't stay open on mobile after clicking icon


mikemacgregor:
I've fussed with this for a while and can't figure it out. When I click the search icon on mobile, the search form from my custom HTML widget opens briefly but immediately closes, making it not usable. What is the trick for it to stay open, or why does it close immediately? Thanks for the help.

https://www.draftbuddy.com/

crftwrk:
The Top Nav Search widget position is made for the default `searchform.php`. If you click on the search button, the collapse opens and searchform gets focused. If searchform loses focus, the collapse hides. You can use any other widgets in this position, but not another saerchform. In your case, there is a searchform that gets not focused by clicking the search button. Collapse opens, checks that form is not focused and closes collapse. 

This is done by a snippet in main-theme https://github.com/bootscore/bootscore/blob/main/js/theme.js in line`21`:

```js
// Close collapse if searchform loses focus
$('.top-nav-search input:first-of-type').on('focusout', function () {
  $('#collapse-search').collapse('hide');
});

```

Delete or comment snippet and let me know if that works. Maybe we can improve this using a snippet that closes collapse on click somewhere outside like this:


```js
// Close collapse if click outside searchform
$(document).on("click", function (event) {
  if ($(event.target).closest("#collapse-search").length === 0) {
    $('#collapse-search').collapse('hide');
  }
});

```

mikemacgregor:
Yes, that seems to have worked. I commented out the first part in theme.js and added the second one to custom.js in my child theme (and included a note about adjusting the parent theme). Thanks very much!

crftwrk:
Do you want to create a PR to change this in main-theme?

PR #411, opened 2023-03-01
Fix searchform collapse behaviour

crftwrk:
This PR improves/fixes the searchform collapse behaviour.

### Before

https://bootscore.me

- Click on search toggler opens collapse and focusses the searchform
- If searchform loses focos, collapse hide

This works fine if bootScore/Woo searchorm or a a completely different widget is used in this position. In https://github.com/bootscore/bootscore/issues/312, a third party searchform is used. The result: On toggler click, searchform collapse opens but form gets **not** focussed. Because there is no focus, collapse close by itself.

### After

Closing collapse is now decoupled from searchform focus. Snippet checks if a click is outside `#collapse-search`and closes collapse then.

- Click on search toggler opens collapse and focusses the searchform
- If click is outside `#collapse-search`, collapse hide

#### Test

https://dev.bootscore.me/

- Click search-toggler in the navbar
- Click inside `#collapse-search` just 1px left or right next to the search´form - searchform stays open
- Click somewhere outside - collapse hides

Closes https://github.com/bootscore/bootscore/issues/312

### DIFF
--- a/js/theme.js
+++ b/js/theme.js
@@ -18,9 +18,11 @@ jQuery(function ($) {
     $('.top-nav-search input:first-of-type').trigger('focus');
   });
 
-  // Close collapse if searchform loses focus
-  $('.top-nav-search input:first-of-type').on('focusout', function () {
-    $('#collapse-search').collapse('hide');
+  // Close collapse if click outside searchform
+  $(document).on('click', function (event) {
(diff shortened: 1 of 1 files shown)


justinkruit:
I know having a click check on _every_ body click could get heavy for a browser, but I don't notice a performance hit on this, so this should alright.

justinkruit (approved):

PR #413, opened 2023-03-03
Add a short code for inserting font awesome signs

kreincke:
As we've discussed in the past: having a short for inserting font awesome symbols instead of having to type the complete html code be an advantage


crftwrk (approved):
👍

PR #414, opened 2023-03-03
Switch from medium featured images to an own bootScore size

kreincke:
There seems to be a weakness concerning dealing with Featured Images. I've described the effect in two posts: https://fodina.de/image-d/ and https://fodina.de/image-e/. If you want to see the difference, use your desktop machine and try this

* Go to https://fodina.de/ and look at the post list. The square featured images are sharp. 
* Insert the word 'Picture' into the search box and look at that list (delivered by the unmodified archive.php). On large Screens, everything blurs.

I am going to describe the technical solution in a longer statement, posted in discussions.

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -35,7 +35,7 @@ get_header();
                 <div class="row">
                   <!-- Featured Image-->
                   <?php if (has_post_thumbnail())
-                    echo '<div class="card-img-left-md col-lg-5">' . get_the_post_thumbnail(null, 'medium') . '</div>';
+                    echo '<div class="card-img-left-md col-lg-5">' . get_the_post_thumbnail(null, 'bsTeaser') . '</div>';
                   ?>
                   <div class="col">
                     <div class="card-body">
--- a/functions.php
+++ b/functions.php
@@ -497,3 +497,8 @@ add_filter('gutenberg_use_widgets_block_editor', '__return_false');
 // Disables the block editor from managing widgets.
 add_filter('use_widgets_block_editor', '__return_false');
 // Disable Gutenberg blocks in widgets (WordPress 5.8) END
+
+// let WordPress automatically created a 'thumbnail' 600x600
+// if a picture is uploaded to the media library
+// Make that size accessible under the name 'bsTeaser'
+add_image_size( 'bsTeaser', 600, 600, true );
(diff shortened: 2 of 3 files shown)


justinkruit:
Personally I feel like this change is more up to the developer, instead of in the theme by default. 

Also what I usually do is call a bigger size, like large, and make it square using the Bootstrap ratios. That way I have more control over the size if it needs to change depending on screen size.

kreincke:
possible. But the more large pictures you have to distribute over the network, the slower the site. 

However, I think that such a valuable work as bootScore should not show blurred and distorted images by default. There's never a second chance for the first impression. Yes, the user could fix it himself, but why do we want to burden him with something that is very little effort on our side:  it costs only 1 line more, and 4 changed values to improve it.

crftwrk:
As @justinkruit said, image sizes should be the developers job. Because we follow straight the WordPress standard and changing default WP image sizes will produce more questions than answers like here https://github.com/bootscore/bootscore/discussions/399.

However, I understand the thing with the super large screens. But:

- Bootstrap is mobile first, means designed for smaller screens.
- `thumbnail` or `medium` are the default WP preview image sizes .
- Changing to larger sizes will load larger images on mobile as well, bad for SEO.
- XXL screens have often a larger screen resolution. Means the image is larger than the original size is. Of course, it's not sharp anymore.

What you can do:

- Use `. get_the_post_thumbnail(null, 'large') . ` instead of `medium`.
- Shorten the excerpt text or write a custom excerpt to lower card (and image) height.
- Use a different loop template https://github.com/bootscore/bs-loop-templates which not crop the featured preview image. For example this one https://bootscore.me/archives/equal-height-sidebar-right/.

kreincke:
Many thanks for your answers. Unfortunately, I cannot agree with you in this case. A tool creating bad optical effects, by default, will perhaps lose its advantage: Again you never get a second chance to make a first impression. If you have templates, which can better deal with the default image sizes of WordPress, then bootScore should perhaps take one of them as 'index.php'

However, my opinion is nothing more than my opinion. Hence, if you don't like my solution, there's no need to integrate this PR. And yes, I will test the other bs-loop-templates later on. 

with best regards KR

crftwrk:
We can generally rethink and improve list view in loop. But let's do this carefully and later.

kreincke:
I've just summarized our discussion: [https://fodina.de/image-i/](https://fodina.de/image-i/) Fell free to contact me, if you think the post contains unfair remarks. I wanted to describe the case, not to blame anyone. Again and again and again: I appreciate your work very much! KR

PR #420, opened 2023-03-04
Remove woo table templates and alert clean-up

crftwrk:

This PR does it with most `rounded` class in templates.

- Removed the last `alert alert-*` and replaces them by `woocommerce-*` classes. Due to https://github.com/bootscore/bootscore/pull/403, all Woo alerts have now an icon, default alerts not.
- Removed most `rounded` classes in templates and replaces tem by `border-radius: $border-radius;`. 
- 6 files that only had changes in the `rounded` class has been deleted
- Minor improvements in the _scss files

This has no impact on the frontend. Tested, works fine and is already live on bootscore.me. Just click through the store and check button classes in cart, checkout and dashboard.


justinkruit (approved):
I'm always up for removing WooCommerce template parts as they're a pain in the butt to maintain. All looks good 👍

PR #428, opened 2023-03-18
Improve horizontal loop cards

crftwrk:
This PR improves the horizontal loop cards in `index.php`, `archive.php` and `content-search.php`. 

Demo https://dev.bootscore.me/, https://dev.bootscore.me/?s=image, https://dev.bootscore.me/category/classic/
Closes https://github.com/bootscore/bootscore/pull/414
Discussion https://github.com/bootscore/bootscore/discussions/426

### Before
before

### After
after

- The featured image (cropped `lg`) has a bit more space, less blur on critical size between `lg` and `xl` breakpoints. Sidebar width is narrower.
- `<h2>` title has now class `h5` like the Bootstrap [horizontal card demo](https://getbootstrap.com/docs/5.3/components/card/#horizontal)
- Badges has now class `text-bg-light` instead `text-bg-secondary` class for less dominance.
- Tags badge function has been refactored to separate "Tagged" title and hide title in the cards.
- Featured image, title, excerpt and read more link to the article. Meta and badges link to tags, category, comments etc. That's why a stretched link to the article is not possible here.

### Change image size by content

This is no guarantee that the image fits on every screen size, because image size depends on card content on the right. It's up to the user to write a short custom excerpt or limit excerpt to less words with a function. Badges and meta can be removed in template files or by CSS. If content is shorten enough, image has nearly original size on >`lg` screens.

tags-removed

meta-removed

### Breaking changes

PR has not a **breaking** change. But users who have overridden page and post templates with sidebar in their child, should change class there from `col-xxl-9` into `col-lg-9` to fit new sidebar width to the article `col`. I'm ok with that.

Cards have to be changed in loop templates and plugins as well.

- [x] bS Grid https://github.com/bootscore/bs-grid/pull/15
- [x] bS Swiper https://github.com/bootscore/bs-swiper/pull/27
- [x] bS Loop Templates https://github.com/bootscore/bs-loop-templates/pull/3
- [x] bS Isotope https://github.com/bootscore/bs-isotope/pull/7

### What else?

Some minor improvements in comments and scss files.

@kreincke I think this should clear it up how the horizontal cards works.


justinkruit (approved):
All looks good 👍

PR #439, opened 2023-04-01
Add bootscore_container_class() function

sweetappleuk:
Allows you to optionally create fluid layouts.

If constant BOOTSCORE_CONTAINER_FLUID is true, the  'container-fluid' class is output in all current templates


sweetappleuk:
The reason for using the constant was so that people would have to very deliberately choose to use fluid layouts. If it's made too "easy", then the project maintainers are likely to get more complaints that the default styling looks odd on fluid layouts and hence creates more work...

But if you'd prefer it's made pluggable I'm happy to do that.

sweetappleuk:
Refactored to be pluggable as requested.

If that's ok will create a pull request for bs-loop-templates

crftwrk (dismissed):
I'm in 👍

2 spaces instead 4 for formatting would be nice

```php
if( ! function_exists('bootscore_container_class') ) {
  function bootscore_container_class() {
    return "container";
  }
}
```

crftwrk (approved):

justinkruit (approved):

crftwrk:
https://bootscore.me/documentation/fluid-layout/

PR #458, opened 2023-04-19
Make main col width dynamic

crftwrk:
Main content `col` must have a fixed width `<div class="col-md-8 col-lg-9">`. Dynamic `col` class isn't enough, because if content has e.g. a larger image `col` uses the entire screen width  and sidebar is pushed to bottom.

This PR replaces the fixed col width classes with function `<div class="<?php echo bootscore_main_col_class(); ?>">`, which returns to `<div class="col-md-8 col-lg-9">` if sidebar is active and to `<div class="col">` if no sidebar is active. In conclusion, if no sidebar is active, content uses entire width instead showing whitespace.

### DIFF
--- a/archive.php
+++ b/archive.php
@@ -17,7 +17,7 @@ get_header();
     <?php bs_after_primary(); ?>
 
     <div class="row">
-      <div class="col">
+      <div class="<?php echo bootscore_main_col_class(); ?>">
 
         <main id="main" class="site-main">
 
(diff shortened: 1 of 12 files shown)


justinkruit:
Just for cleanness, do we want to use `<?php echo [] ?>` or `<?= [] ?>`? My personal preference is always `<?=` because it looks cleaner to me and is still a clear indication something is echoing.

crftwrk:
Your ideas are always better than mine. Want to change that quickly?

justinkruit:
Sure! I'll do it tomorrow though, going out for the evening 😉

Issue #460, opened 2023-04-22
Merge both headers into one and enable Woo parts if WooCommerce is enabled


crftwrk:
### Describe the feature that you'd like

We can use `get_template_parts()` to add user and cart into `header.php` and get enabled if WooCommerce plugin is installed and activated.

In v6 we can do the same with `woocommerce-functions.php` and `_bscore_woocommerce.scss`. So, full control is back to the main-theme. `bootcommerce-child` is obsolete and can be deleted or moved to archive. `bootscore-child` is now the only child-theme and can manage a shop site as well.

#### v5

- [ ] Use `get_template_parts()` for WooCommerce header parts and enable them if WooCommerce is active
- [ ] Delete `header-woocommerce.php`
- [ ] Delete `header.php` in bootcommerce-child

#### v6

- [ ] Enable `woocommerce-functions.php` if WooCommerce is activated
- [ ] Find a way to compile `_bscore_woocommerce.scss` if WooCommerce is activated
- [ ] Delete or archive bootcommerce-child theme

### Why is this feature needed? What problem does it solve?

- We have to maintain only one `header.php` instead three.
- We have to maintain only one child-theme instead two.
- More control to the main-theme. All changes are only done there.
- Main-theme can be used as a stand-alone shop without touching the code.

### Checklist

- [X] I have searched the issues and this feature has not been requested.

PR #461, opened 2023-04-22
Refactor header.php and enable Woo parts if WooCommerce is installed

crftwrk:
In relation to https://github.com/bootscore/bootscore/issues/460

This PR uses `get_template_part` to get header buttons/widgets. WooCommerce related items load when WooCommerce is installed. No 2 different `header.php`'s needed anymore.

- No breaking changes to existing bootcommerce-child installations
- Allows us to delete `header.php` in bootcommerce-child

## How to test

Test in main-theme

- Remove `//` comments in `main.scss` line `4` to enable Woo styles
- Add following snippet to `functions.php` to enable Woo scripts if Woo is installed

```php
if ( class_exists( 'WooCommerce' ) ) {

  require get_template_directory() . '/woocommerce/woocommerce-functions.php';
  
  function my_function() {
    echo '<p class="alert alert-success">Woo installed</p>';
  }
  add_action('bs_after_primary', 'my_function', 5);

} else {

function my_function_2() {
    echo '<p class="alert alert-warning">Woo NOT installed</p>';
  }
  add_action('bs_after_primary', 'my_function_2', 5);

}
```

- Enable/disable WooCommerce plugin in backend and check the header in frontend
- Install a usual bootscore-child and override parts or entire header templates

@justinkruit 
- If you like it, merge it
- Is there a way to add `_bscore_woocommerce.scss` to the compiler and enable file only if WooCommerce is installed?


justinkruit:
Maybe an idea to move the files a bit? Instead of them being named `header-top-nav-search-collapse-woocommerce.php`, name them `header/top-nav-search-collapse-woocommerce.php`, and that way you can call this one via `get_template_part('template-parts/header/top-nav-search-collapse', 'woocommerce');` (as woocommerce is a variant of that file)

crftwrk:
Good idea, pls check. I removed some comments because snippets speak by itself.

justinkruit (approved):
Looks great, good job on this 👍

PR #465, opened 2023-04-28
Fix bootscore_category() failing to output categories

sweetappleuk:
printf() missing a format string...


crftwrk (approved):
Thank you for figuring this out 👍

However, I do not know why this function is still there because this was replaced by category badges a long time ago.

PR #473, opened 2023-05-08
Enqueue scripts if WooCommerce is installed

crftwrk:
- Closes https://github.com/bootscore/bootscore/issues/460
- Related PR https://github.com/bootscore/bootscore-child/pull/27/files

This PR enables Woo scripts if plugin is installed and has no breaking changes for existing bootcommerce-child installations.

- `woocommerce-functions.php` content has been moved to `wc-functions.php`
- `woocommerce-functions.php` still exists because existing bootcommerce-child installations have `require get_template_directory() . '/woocommerce/woocommerce-functions.php';` in their `functions.php`
- Styles for WooCommerce gets compiled by default. Had a look to it, this are round about 700 beautified lines of compiled CSS. There is a comment that `@import` line can be deleted if Woo styles are not needed. Related PR for bootscore-child theme https://github.com/bootscore/bootscore-child/pull/27/files
- When merged, we can archive or close the bootcommerce-child repo, because main-theme and bootscore-child are now be able to manage a shop. One repo less.

Tested it on some live shops and works fine.


justinkruit (approved):
This all looks great to me, and a next step to make updates easier. 👍

PR #487, opened 2023-05-31
Update Bootstrap 5.3.0

crftwrk:
This PR has a lot changed files, but it must not reviewed each of them. Checking the overall functionality is enough and has no breaking changes to existing installations.

Demo https://dev.bootscore.me (dark-mode switcher is in the footer)

- Update to Bootstrap 5.3
- Changing color classes to new one, `bg-light` > `bg-body-tertiary`, `text-muted` > `text-body-tertiary` etc.
- Adding a fallback to old color classes
- Update versions, readme.txt and readme.md
- Some minor bugfixes and improvements, nothing special

Tested it on some live-sites and works fine.

There is another PR https://github.com/bootscore/bootscore/pull/483. This updates two templates for the upcoming WooCommerce 7.8. Tested both templates with current 7.7 and this has no issues. So, I think we can merge this as well. This will prevent us to create a new release in two weeks again.

@justinkruit if you like it, merge it! As usual, there is a drafted release ready. bootscore.me site is ready as well.

Edit: From here, I think we should decouple versions from Bootstrap versions, switch to a 3 digit number and take semantic versioning serious as criticized  in some issues.

Closes https://github.com/bootscore/bootscore/issues/489
Closes https://github.com/bootscore/bootscore/issues/488
Closes https://github.com/bootscore/bootscore/issues/490


justinkruit:
@crftwrk There is a merge conflict in this one, could you take a look at that?

Also, I agree on the semantic versioning. Do we want to start that with us beginning at 5.3.0? Or do we want to start a bit fresher, make it something like v6, and go from there? Not all that good with versioning often 😅 😂

crftwrk:
Yes, I check it right now

justinkruit (approved):

crftwrk:
It was just the Stripe field for the dark mode. But it's good as it is. 

> Do we want to start that with us beginning at 5.3.0?

Yes, I think so. Next minor update will be 5.3.1, next major 5.4.0 and so on. No matter what Bootstrap version says. If Bootstrap is v6, we can also go to 6.x.x. This gives us better control over the features and provides a clear number to the users. Agree?

justinkruit:
I agree.

I'm currently making some very small changes, and will maybe do another round of formatting just to be sure that everything is nice for 5.3.

crftwrk:
Perfect

PR #505, opened 2023-06-12
Reduce woo templates 2

crftwrk:
This PR removes 4 more WC templates and a slightly design update. Not every file must be reviewed in detail, overall functionality is enough.

- sale-flash uses the default WC `Sale!` text instead `%` https://dev.bootscore.me/shop/
- Card has been removed in form-login, think this is a bit cleaner https://dev.bootscore.me/my-account/. SCSS code for card remains commented, so we can quickly readd the card https://github.com/bootscore/bootscore/pull/505/files#diff-c4072331985615befc736936206ce7c7e785b9e590b77e1b74ba3faf729f630e.
- Same in checkout form-login https://dev.bootscore.me/checkout/
- Horizontal forms have a smaller gap, looks a bit more like `g-4` class. 

| Before  | After |
| login before  | login-after  |
| checkout before  | checkout after  |
| sale before  | sale after  |

Once all WC files https://github.com/bootscore/bootscore/issues/333 have ben reviewed/removed, I will do a general SCSS clean-up and breakdown SCSS to WC components. For now, SCSS files are a bit chaotic organised but work well.

@justinkruit if you like it, merge it!


justinkruit (approved):
All looks good to me 👍

Issue #511, opened 2023-06-21
top-nav-widget ms


exlexv:
### Describe the feature that you'd like

Hello & thanks for Bootscore:)

Most buttons in header are ms-1 ms-md-2 (togglers, cart, user etc.). But top-nav-widget is registered with ms-2, so you have to edit parent theme for "ms-1 ms-md-2" or copy entire widget function or add css  @media (max-width: 768px) {.top-nav-widget {margin-left: 0.25rem!important;}} . Maybe change it initially to "ms-1 ms-md-2" :)? 

  `function bootscore_widgets_init() {`
   ` // Top Nav`
 `...`
     ` 'before_widget' => '<div class="top-nav-widget ms-2 ">',`

### Checklist

- [X] I have searched the issues and this feature has not been requested.

crftwrk:
Thank you for your request, but may I ask what's your case for this? `top- nav` widget position is next to the logo on mobile and `ms-2` class is used to get space to the expanded menu. Changing to `ms-1 ms-md-2` will have no effect?

exlexv:
top-nav widget as i've just checked is next to mobile toggle on mobile and the same on desktop. So when for example other buttons added top top-nav, they look ok on desktop but not on moobile.

crftwrk:
On mobile there is nothing on the left side before this widget position, only the logo https://dev.bootscore.me/. 

If you think this is a usefull improvement, please fork the theme and create a pull request. We will merge it.

exlexv:
Yes, nothing on the left, only logo, and `serch` on the right. But if you're going to add some additional buttons to the widget. Let's say we are adding 2 more buttons via wordpress admin panel, one after another in 2 different `HTML-code` containers. And the result on mobile would be like on img [https://omfm.ru/123.png](url) But with `ms-1 ms-md-2` everything's gonna be ok:)

crftwrk:
Understand. Please create a pull request, we will merge it.

PR #512, opened 2023-06-21
Changing ms-2 to ms-1 ms-md-2 in top-nav-widget

exlexv:
Following  #511 
Changing `ms-2` to `ms-1 ms-md-2` in top-nav-widget. So if we add items or buttons via `wordpress admin panel` to the top-nav widget, one after another in different `HTML-code` container blocks, we would keep 0.25rem margin between them on mobile screens instead of ms-2, which would look too wide compared to margin between other elements (navbar-toggler, search, cart, user — all of them use ms-1 on mobile). So we won't see this:)

Closes https://github.com/bootscore/bootscore/issues/511


crftwrk (approved):

crftwrk:
Thank you!

PR #520, opened 2023-06-28
Reduce woo templates 4

crftwrk:
This PR removes 2 more WC templates in the checkout and some minor improvements. Tested on some live-shops and works fine. Not every file must be reviewed in detail, overall functionality is enough.

Demo: https://dev.bootscore.me/shop/ add something to the cart and go to checkout

- Deleted `form-billing.php`. The function to add `form-control` class to checkout billing form is not needed anymore and has been deleted as well.
- Deleted `form-shipping.php`. Checkboxes are now done by SCSS.

What else?
- Replaces some overlooked `<? php echo` with shorthand
- Removes the `.nav-link:focus-visible` box-shadow
- Wraps the `<h1>` in full-width-image templates in a `<div>`. Closes https://github.com/bootscore/bootscore/issues/519

@justinkruit if you like it, merge it!

PR #525, opened 2023-07-07
Update WooCommerce 7.9 templates

crftwrk:
This PR updates `cart.php` and `mini-cart.php` for upcoming WooCommerce 7.9 (July 11th) release. I hate maintaining WC templates, so this two files are the next ones on my removing list ;-)

Tested it with WC 7.8.2 as well and works fine, we can safely merge.

@justinkruit think we should https://github.com/bootscore/bootscore/pull/520 first. If both PR's are merged, I will prepare a release for 5.3.1 which we should ship on Monday. Agree?

### DIFF
--- a/woocommerce/cart/cart.php
+++ b/woocommerce/cart/cart.php
@@ -13,7 +13,7 @@
  *
  * @see     https://docs.woocommerce.com/document/template-structure/
  * @package WooCommerce\Templates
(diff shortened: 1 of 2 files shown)


justinkruit (approved):
Haha, I agree, I also don't like updating templates.
I should really update our theme at work soon 😅

PR #534, opened 2023-07-26
Update Bootstrap 5.3.1

crftwrk:
Tested and works fine. But we should wait 2 weeks for the next release because WC 8.0 is in the pipe and I'm not sure if they will do changes in the templates.


justinkruit (approved):
👍 

Anything special that will be happening in WC 8.0?

crftwrk:
> Anything special that will be happening in WC 8.0?

Not sure. Have Beta-2 installed and still no template changes, but in the past they did changed in the RC. Release date is 8th August.

PR #535, opened 2023-07-29
Reduce Woo templates 5

crftwrk:
This PR removes `cart.php`, `breadcrumb.php` and has minor improvements. Icons are replaced with mask icons. So, we can ship a bit more than only updated Bootstrap in next release. Tested, works fine and is already live on bootscore.me.

#### One weird thing 

There was an overlooked snippet in the `cart.php` which have been changed by Woo in one of the past updates. This snippet affects:

https://dev.bootscore.me/shop/

- Beanie, 1 left in stock
- Hoodie with pocket, sold individually

In cart page before, the quantity was replaced by number "1" for individually sold products and disabled form for only 1 in stock products. Now quantity cell is empty. 

If only 1 in stock, quantity is hidden in single product page.

I triple checked it with default WordPress themes and this is what WooCommerce exactly do in all themes (check screenshots below). 

| Before  | After |
| cart before  | cart after  |
| individually before  | individually after  |
| 1 in stock before  | 1 in stock after  |

| Default theme  | Default theme |
| cart default  | individually default  |
| 1 in stock default  |   |


justinkruit:
Seems like the change to the cart quantity happened here: https://github.com/woocommerce/woocommerce/pull/36460

I'm personally not a huge fan of the quantity being hidden... Do we want to change this behavior in bootScore, or keep it this way and let the devs figure it out?

crftwrk:
> I'm personally not a huge fan of the quantity being hidden... Do we want to change this behavior

Totally agree. Do you have an idea without editing the template file?

justinkruit:
I don't see the benefit of making the input a text type instead of number. Also, the snippet of the PR shows the problem that every type after that in the cart also has the type text so you need to remove the filter again too. I think just having it be the number type should be enough, right?

crftwrk:
Whats about this?

```php
/**
 * This snippet can be used to force the quantity input to display in cases where
 * the input value cannot be changed (which is when the product is set to be sold
 * individually, or when the min and max values are identical).
 */
add_filter('woocommerce_quantity_input_args', function (array $args) {

	if ($args['max_value'] < 1 || $args['min_value'] !== $args['max_value']) {
		remove_filter('woocommerce_quantity_input_type', 'change_quantity_input_type');
		return $args;
	}
	add_filter('woocommerce_quantity_input_type', 'change_quantity_input_type');
	$args['readonly'] = true;
	return $args;
});

function change_quantity_input_type() {
	return 'text';
}

```

justinkruit:
That still sets the type to text tho? 🙈

crftwrk:
We can simply do 

```php
function change_quantity_input_type() {
	return 'number';
}
```

Seems to work well. I have personally no problems to show the qty input everywhere. It's disabled for individually sold and 1 in stock products. For me it's clear.

However, if we are not sure, we can open a new PR for this and decide later if we merge or not?

justinkruit:
I think having the input always shown but disabled when editing is not possible, would be the right decision.
Maybe good to indeed make a new PR, but also accept it immediately (just for PR clarity).

crftwrk:
When merged, I do a quick PR immediately.

justinkruit (approved):
👍

Issue #538, opened 2023-08-06
Revisit qty in cart


crftwrk:
https://github.com/bootscore/bootscore/pull/535

PR #543, opened 2023-08-10
Make qty always visible

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/538

Demo https://dev.bootscore.me/shop/

As discussed in https://github.com/bootscore/bootscore/pull/535, this PR makes the qty always visible and fixes overflow text in mobile cart coupon button for languages with longer words.

| Before | After |
| Default product  | No changes  |
| 1 in stock qty hidden  | 1 in stock qty visible but disabled  |
| Sold individually qty hidden  | Sold individually qty visible but disabled  |
| Cart 1 in stock qty visible, can be set to 0  | No changes  |
| Cart sold individually qty hidden  | Cart sold individually qty visible, can NOT be set to 0  |

| Before  | After |
| default-before  | default-after  |
| one-ins tock-before  | one-in-stock-after  |
| sold-individually-before  | sold-individually-after  |
| cart-before  | cart-after  |
| cart-mobile-before  | cart-mobile-after  |

Both products, 1 in stock and sold individually are not handled by WooCommerce in the same way. So, qty behaviour in cart is different. But both products are mostly edge cases.

I personally prefer to have the qty always visible and disabled if not more than one product can be added to the cart. But I am fine as well as we leave it as it is and hide qty in cart for sold individually products. 

@justinkruit if you like it, merge it. If you dislike it, close it. Think we should ship this with 5.3.2 https://github.com/bootscore/bootscore/pull/541 

Edit: Added a `:hover` state to WC breadcrumb home icon

### DIFF
--- a/scss/bootscore_woocommerce/_wc_breadcrumb.scss
+++ b/scss/bootscore_woocommerce/_wc_breadcrumb.scss
@@ -5,25 +5,32 @@ WooCommerce Breadcrumb
 .wc-breadcrumb {
 
   // Add a Fontawesome house icon for empty "Home" (wc-functions.php)
-  .breadcrumb-item:first-child a::before {
-    content: ' ';
-    display: inline-block;
-    width: $font-size-base;
-    height: 100%;
-    background-color: var(--#{$prefix}link-color);
-    mask-position: center;
-    mask-repeat: no-repeat;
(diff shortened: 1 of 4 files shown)


justinkruit:
Do we want to add this to v5.3.2?

crftwrk:
Why not?

justinkruit:
> Why not?

I support it, what's why I'm asking. Didn't see it was already included for the v5.3.2 release, my bad.

justinkruit (approved):

PR #545, opened 2023-08-12
Split functions.php into smaller files

crftwrk:
This PR cleans up the patchwork mess in `functions.php` and splits functions into smaller files in  `inc` folder. `functions.php` uses

```php
require_once('inc/breadcrumb.php');
...
```

to include them.

Files have now deny direct access, which we should add to all php files step-by-step:

```php
defined( 'ABSPATH' ) || exit;
```

- Comments have been updated and follows the standard
- Files have now a version to track each file changes

Tested and works fine. If this PR is accepted and merged, I'll do the same with `wc-functions.php`.

@justinkruit if you like it, merge it!


justinkruit (approved):
Looks good!

Also, is there a reason we use the `if() : endif;` instead of `{}`?

PR #555, opened 2023-08-28
Split wc-functions into smaller files

crftwrk:
Same as https://github.com/bootscore/bootscore/pull/545. Tested and works fine.

@justinkruit if you like it, merge it ;-)


woocommerce/inc/wc-enqueue.php
```diff
@@ -0,0 +1,35 @@
...
+  
+  // Enable fragments script (disabled in WooCommerce 7.8.0 if mini-cart widget is not in a widget position)
+  // See https://developer.woocommerce.com/2023/05/24/woocommerce-7-8-beta-1-released/
+  wp_enqueue_script('wc-cart-fragments');
+
+  // Get modification time. Enqueue files with modification date to prevent browser from loading cached scripts and styles when file content changes. 
+  $modificated_WooCommerceJS = date('YmdHi', filemtime(get_template_directory() . '/woocommerce/js/woocommerce.js'));
+
+  // WooCommerce JS
+  wp_enqueue_script('woocommerce-script', get_template_directory_uri() . '/woocommerce/js/woocommerce.js', array(), $modificated_WooCommerceJS, true);
```
justinkruit, line 28: Same as the function, adding bootscore to the handle could keep it clear it's from the theme when checking loaded resources.

justinkruit (commented):
Got 2 comments, other than that I like the clarity it provides by dividing it👍

woocommerce/inc/wc-enqueue.php
```diff
@@ -0,0 +1,35 @@
...
+
+
+// Exit if accessed directly
+defined( 'ABSPATH' ) || exit;
+
+
+/**
+ * Enqueue scripts
+ */
+function wc_scripts() {
```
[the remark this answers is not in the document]
crftwrk: Done

woocommerce/inc/wc-enqueue.php
crftwrk: What do you mean exactly here?

crftwrk: Got it, done.

https://github.com/bootscore/bootscore/pull/555/commits/d4f5895e1cbdb9bb19c92c60d979a3ed1293a6f6

justinkruit (approved):
With the new changes, seems good to me 👍

Issue #343, opened 2023-01-03
Refactor Woo checkout form validation


crftwrk:
```scss
.woocommerce-invalid .form-control {
  @extend .is-invalid;
}

.woocommerce-validated .form-control {
  @extend .is-valid;
}

```

Much more better than

```js
// Checkout Form Validation
$('body').on('blur change', '.form-row input', function () {
  $('.woocommerce form .form-row.woocommerce-validated .select2-container, .woocommerce form .form-row.woocommerce-validated input.input-text, .woocommerce form .form-row.woocommerce-validated select, .woocommerce form .form-row.woocommerce-validated .form-check-input[type=checkbox]').removeClass('is-invalid').addClass('is-valid');
  $('.woocommerce form .form-row.woocommerce-invalid .select2-container, .woocommerce form .form-row.woocommerce-invalid input.input-text, .woocommerce form .form-row.woocommerce-invalid select, .woocommerce form .form-row.woocommerce-invalid .form-check-input[type=checkbox]').removeClass('is-valid').addClass('is-invalid');
});

```

We can do this after https://github.com/bootscore/bootscore/pull/329 is merged

crftwrk:
We leave it as it is, checks are too messy https://github.com/bootscore/bootscore/commit/daa59819101e897de63eca8096d8efb36496cf42

PR #557, opened 2023-08-30
Reduce WC templates 7

crftwrk:
This PR removes 3 more WC templates in checkout according to WC 8.1 which comes on Tuesday next week https://developer.woocommerce.com/2023/08/29/woocommerce-8-1-0-beta-1-released/ and improves the form validation. Tested, works fine and is already live on bootscore.me. Not each file must be reviewed, the overall check is enough.

Checkout shortcode will be replaced by checkout block in WooCommerce in November https://developer.woocommerce.com/2023/08/18/cart-and-checkout-blocks-becoming-the-default-experience/. So, we can delete file in v6.

| Before  | After |
| before  | after  |

@justinkruit if you like it, merge it ;-)

Closes https://github.com/bootscore/bootscore/issues/343


justinkruit:
> Checkout shortcode will be replaced by checkout block in WooCommerce in November https://developer.woocommerce.com/2023/08/18/cart-and-checkout-blocks-becoming-the-default-experience/. So, we can delete file in v6.

I'm having my doubts about that. I like the possibility of changing the layout, and I think a lot of developers do so. As far as my research goes, the blocks are way more limited to code editing. So my opinion is to keep the shortcode edits we've made.

justinkruit:
> Totally agree, but I think we should provide support to the Blocks step-by-step. It seems to be not an impossible thing.

Also agree on that, but maybe keep both supported? The shortcode is far from deprecated, it's just not the default anymore. I think it's best to keep the support in the theme, and focus on how to style both the shortcodes and blocks together.

crftwrk:
> Also agree on that, but maybe keep both supported?

You're right

justinkruit:
Then we agree on that 😉

justinkruit (approved):
Then I'll now stop keeping this PR hostage 😂
All seems good to me 👍

PR #560, opened 2023-09-04
Use function for sidebar col width and responsive offcanvas

crftwrk:
Think this PR speaks for itself. Change the entire main/sidebar columns and breakpoints in one file.

@justinkruit if you like it, merge it. Are you able to publish a release next Monday?

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -19,7 +19,7 @@ defined( 'ABSPATH' ) || exit;
  */
 require_once('inc/theme-setup.php');             // Theme setup and custom theme supports
 require_once('inc/breadcrumb.php');              // Breadcrumb
-require_once('inc/columns.php');                 // Main column width
+require_once('inc/columns.php');                 // Main/sidebar column width and breakpoints
 require_once('inc/comments.php');                // Comments
 require_once('inc/container.php');               // Container class
 require_once('inc/enable-html.php');             // Enable HTML in category and author description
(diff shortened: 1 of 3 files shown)


justinkruit:
Maybe instead of functions, we should use filters/actions?

justinkruit:
Sure! Just quickly wrote this for the sidebar.php

```
<div class="<?= apply_filters('bootscore_sidebar_cols', 'col-md-4 col-lg-3'); ?> order-first order-md-last">
```

With this people can do the following in their child theme:
```php
add_filter('bootscore_sidebar_cols', 'bootscore_sidebar_cols_function');
function bootscore_sidebar_cols_function($string) {
  return 'col-lg-4';
}
```

Which will then override `col-md-4 col-lg-3`.

Maybe we should do this too with the containers?

crftwrk:
The thing is that we **_have_** to do this with containers and main-col as well. Means that we must edit all files again, rewrite the docs and keep existing functions in the trash file to not create breaking changes. I'm fine with the functions right now and we can adress this to v6?

justinkruit:
Sure, we can make a big breaking change for this with v6. Seems like a good plan.

justinkruit (approved):
Looks fine to me.
Honestly looking forward to possibly change everything with filters and make an even more streamlined process 😁

PR #565, opened 2023-09-08
Rewrite theme.js to remove jQuery

androidacy-user:
jQuery quite frankly is a burdensome and weighty library in today's ES6 world. This file is one of the only files in bootscore that still depends on jQuery to function.


justinkruit:
> This file is one of the only files in bootscore that still depends on jQuery to function.

That's actually not completely true, there is also [woocommerce/js/woocommerce.js](https://github.com/bootscore/bootscore/blob/5cc2a71e43e9296710acf0a092f62cb0e9adac5b/woocommerce/js/woocommerce.js) and the insides of [woocommerce/inc/ajax-add-to-cart.php](https://github.com/bootscore/bootscore/blob/5cc2a71e43e9296710acf0a092f62cb0e9adac5b/woocommerce/inc/ajax-add-to-cart.php). So these changes will not completely remove jQuery as dependency, at least not when using woocommerce.

crftwrk:
I had this on my list as well. It's good that already someone want to do this. Tested it and found 2 issues. Readded the jQuery again to show how this snippets should work:

1. First snippet `// Handle offcanvas links click event.` does not work. Check https://bootscore.me/theme/scrollspy-onepager/, shrink browser to mobile view and click the links in menu. Offcanvas closes if one of the onepager-links is clicked. Snippet can be deleted if https://github.com/twbs/bootstrap/pull/38514 is merged.
2. Fourth snippet `// Hide collapse-search when clicking outside of it.` Does not work as well. Click on the same page the search-toggler in the navbar. Click 1px right next to the input-group search button - collapse stays open. Click somewhere else on the whitespace - collapse closes.

About the other 2 files @justinkruit mentioned. I'm a huge friend to have a standard and Vanilla JS there would be nice as well. But both files gets only enqueued if WooCommerce plugin is installed and entire WooCommerce' JS is jQuery. It will not make a huge different if we change them to Vanilla JS too. So, in this case I'm fine using Vanilla in theme.js and jQuery in WC scripts. 

So users who haven't installed WooCommerce have a theme without jQuery dependency when this is merged.

Just for info:
> I'm not entirely comfortable rewriting those files to remove jquery as we cannot test it

Simply setup a testing site, install WooCommerce and you can test them well. But as I wrote above, I think this is not important for now.

And while we are here, Internet Explorer warning seems to be outdated and many other browsers are not supported by Bootstrap 5 as well. Think we can delete this in a further update.

If we get those 2 issues fixed, I'm in. Let me know what you think.

justinkruit:
And when all is done and fixed, please request both me and @crftwrk as reviewer (if you're able to, not entirely sure if you can as PR author so please test it).

I'd like to fully test this when the bugs mentioned above are fixed.

crftwrk:
Set this to draft to avoid accidentally merging.

crftwrk:
Nearly ;-)

- I miss the comments and have to read the code first to understand what each snippet does.
- Self closing offcanvas by click on `<a>` does still not work
- NOT tested IE warning, because my old machine has gone. But as mentioned above, think we should remove this snippet in a separate PR.
- All other snippets works fine
- All snippets are wrapped in `function initializeScripts() {`. Why?

Changed some code and everything (IE warning not tested but not important) works fine. Is already live on bootscore.me.


```js
/*--------------------------------------------------------------
Theme JS
--------------------------------------------------------------*/

// Close offcanvas on click a, keep .dropdown-menu open 
// See https://github.com/bootscore/bootscore/discussions/347
document.addEventListener('DOMContentLoaded', function () { // Fire after bootstrap.bundle.min.js is loaded 
  const offcanvasLinks = document.querySelectorAll('.offcanvas a:not(.dropdown-toggle):not(.remove_from_cart_button)');
  const offcanvas = new bootstrap.Offcanvas(document.querySelector('.offcanvas'));

  offcanvasLinks.forEach(function (link) {
    link.addEventListener('click', function () {
      offcanvas.hide();
    });
  });
});


// Remove search collapse button hide if widget is empty
// Deprecated v5.2.3.4, done by php if (is_active_sidebar('top-nav-search')) in header.php
// Remove in v6
const collapseSearch = document.getElementById('collapse-search');
if (collapseSearch && collapseSearch.children.length === 0) {
  document.querySelectorAll('.top-nav-search-md, .top-nav-search-lg').forEach((element) => element.remove());
}


// Focus searchform on click search-toggler
document.querySelector('#collapse-search').addEventListener('shown.bs.collapse', function () {
  document.querySelector('.top-nav-search input:first-of-type').focus();
});


// Close search collapse if click outside the searchform
document.addEventListener('click', function (event) {
  const target = event.target;
  const collapseSearch = document.querySelector('#collapse-search');
  const collapseButton = document.querySelector('[data-bs-target="#collapse-search"]');

  if (!collapseSearch.contains(target) && collapseSearch.classList.contains('show')) {
    collapseButton.click(); // Simulate a click on the collapse button to hide the collapse
  }
});


// To-top button
const topButton = document.querySelector('.top-button');
if (topButton) {
  window.addEventListener('scroll', function () {
    topButton.classList.toggle('visible', window.scrollY >= 500);
  });
}


// Height classes, add class to your content
['height-50', 'height-75', 'height-85', 'height-100'].forEach((cls) => {
  const factor = parseFloat(cls.split('-')[1]) / 100;
  document.querySelectorAll(`.${cls}`).forEach((element) => {
    element.style.height = `${window.innerHeight * factor}px`;
  });
});


// IE warning
if (window.document.documentMode && typeof bootscore !== 'undefined') {
  const IEWarningDiv = document.createElement('div');
  IEWarningDiv.className = 'position-fixed top-0 end-0 bottom-0 start-0 d-flex justify-content-center align-items-center';
  IEWarningDiv.style.backgroundColor = 'white';
  IEWarningDiv.style.zIndex = 1999;
  IEWarningDiv.innerHTML = `
        <div style="max-width: 90vw;">
            <h1>${bootscore.ie_title}</h1>
            <p class="lead">${bootscore.ie_limited_functionality}</p>
            <p class="lead">${bootscore.ie_modern_browsers_1}${bootscore.ie_modern_browsers_2}${bootscore.ie_modern_browsers_3}${bootscore.ie_modern_browsers_4}${bootscore.ie_modern_browsers_5}</p>
        </div>`;
  document.body.appendChild(IEWarningDiv);
}

```

### How to test
- Go to https://bootscore.me/theme/scrollspy-onepager/. Each section should have 100% window height.
- Shrink browser on the same page to mobile view, open offcanvas and click a nav-link. Offcanvas closes.
- Click the search-toggler in the navbar. Searchform gets focused.
- Click 1px left or right next to the searchform. Searchform loses focus but collapse stays open.
- Click somewhere else on the page. Searchform collapse closes.
- Add an outdated [header.php](https://github.com/bootscore/bootscore/blob/7685ca4b860e1cb4104b9b51085f61091b811652/header.php) to your child. Now go in backend to widgets and remove the searchform in `top-nav-search` area. Search toggler should be removed in frontend.

If you want to change this, I'm in 👍

crftwrk:
The thing is that each snippet has a reason why it's there and has a job to do. If one of this is not working, it may create issues on existing installations. I recommend to setup a test environment instead doing changes on your live site.

Agree that we can skip to test the IE warning, but all other snippets have to work properly. Comments are mandatory as well to help other contributors to understand what each snippet does.

Added in https://github.com/bootscore/bootscore/pull/565#issuecomment-1772451969 a working script for the self-closing offcanvas.

Set this to draft again to avoid accidentally merging. If we can fix that, I'm in 👍

crftwrk:
@androidacy-user FYI, IE warning has been removed https://github.com/bootscore/bootscore/pull/620. You can skip this snippet.

crftwrk:
This comes in v7 https://github.com/bootscore/bootscore/pull/1248

- 100% performance score on desktop https://pagespeed.web.dev/analysis/https-v7-dev-bootscore-me/4w8v9ekxfe?form_factor=desktop
- Woo JS stays in jQuery (for now)

Issue #367, opened 2023-01-18
Deny direct access to theme PHP files?


crftwrk:
### Discussed in https://github.com/bootscore/bootscore/discussions/366

<div type='discussions-op-text'>

<sup>Originally posted by **jmcintyresage** January 18, 2023</sup>
Shortly after public launch, one of our Bootscore-based sites started getting hit with requests for the base theme's index.php and search.php files, resulting in fatal PHP errors because get_header() wasn't defined.

Would the developers consider adding code that would deny direct access to the base theme's PHP files? Something as simple as:

```
// Stop if this file is called directly
defined( 'ABSPATH' ) || exit;
```

Respectfully,
JIm
</div>

#### Theme

- [x] Functions https://github.com/bootscore/bootscore/pull/545
- [x] WC functions https://github.com/bootscore/bootscore/pull/555
- [x] Sidebar https://github.com/bootscore/bootscore/pull/560
- [x] Header https://github.com/bootscore/bootscore/commit/4fe913fa6b4b7a233fdd6095ce508d0d852ca3b6
- [x] Footer https://github.com/bootscore/bootscore/commit/07dd0a2b8af292608d799f71c36b2a88c3d0a4df
- [x] Child https://github.com/bootscore/bootscore-child/commit/3407e9f806761bb6e9c402a4cabf00c4e9497ce1
- [x] bS Loop templates https://github.com/bootscore/bs-loop-templates/pull/9
- [x] Post templates https://github.com/bootscore/bootscore/pull/696
- [x] Page templates https://github.com/bootscore/bootscore/pull/696
- [x] Template parts https://github.com/bootscore/bootscore/pull/696
- [x] Archive https://github.com/bootscore/bootscore/pull/696
- [x] Search https://github.com/bootscore/bootscore/pull/696

##### template-parts
- [x] content-none https://github.com/bootscore/bootscore/pull/602
- [x] content-search https://github.com/bootscore/bootscore/pull/602
- [x] actions-woocommerce https://github.com/bootscore/bootscore/pull/602
- [x] actions https://github.com/bootscore/bootscore/pull/602
- [x] offcanvas-woocommerce https://github.com/bootscore/bootscore/pull/602
- [x] top-nav-search-collapse-woocommerce https://github.com/bootscore/bootscore/pull/696
- [x] top-nav-search-collapse https://github.com/bootscore/bootscore/pull/696

#### Plugins

- [x] bS Isotope https://github.com/bootscore/bs-isotope/pull/31
[...]

PR #575, opened 2023-09-15
Filter refactor

### LINKED ISSUE #566: [Feature Request] Filter to remove fontawesome
### Describe the feature that you'd like

Hello!

Since we use FontAwesome Pro, we remove the default FA styles provided by this theme. However, currently we have to manually edit the functions.php from the parent theme, as trying to override it in the child theme is hit and miss and requires us to mess with priorities which causes other conflicts anyway.

We'd like to see a filter (i.e., "bootscore_fontawesome_enqueue") or similar that can prevent this arguably non-essential file from being loaded.

### Why is this feature needed? What problem does it solve?

Some people like us may prefer to use their own version of fontawesome or even remove it entirely.

### Checklist

- [X] I have searched the issues and this feature has not been requested.

justinkruit:
Adding a lot of filters to replace the small functions with.
This will make it easier to make changes for other developers.

- Turned `bootscore_sidebar_col_class` into a filter (6d7e18a)
- Turned `bootscore_sidebar_toggler_class` into a filter (6d7e18a)
- Turned `bootscore_sidebar_offcanvas_class` into a filter (6d7e18a)
- Turned `bootscore_container_class` into a filter (afb590d c1d8393 d01bd42 1d6c386)
  - Each reference of this filter has an argument included of which page it is. This way you can filter based on that when adding filters to your own code.
  An example is included below.
- Turned `bootscore_main_col_class` into a filter (94239fc)
- Added a filter to the SCSS compiler if the user wants to make changes to `$compiler` before the compile happens (e903f0d)
- Added a filter to make Font Awesome toggleable (4ebdd63) (Fixes #566)
- All the old class functions are being properly handled in deprecated.php (f1d8939)

### Examples
Change the container of the header to `container-fluid`:
```php
function container_class($string, $location) {
  if ($location == 'header') {
    return "container-fluid";
  }
  return $string;
}

add_filter('bootscore/container_class', 'container_class', 10, 2);
```

Turn off Font Awesome
```php
add_filter('bootscore/load_fontawesome', '__return_false');
```


crftwrk:
Cool, I'm in. Outdated scripts we have to put to https://github.com/bootscore/bootscore/blob/main/inc/deprecated.php to keep functionality for overridden templates.

While we are here, do we want include deny direct access and version as described in https://github.com/bootscore/bootscore/issues/367 in this PR? Version should be 5.3.4.

https://github.com/bootscore/bootscore/blob/00a418bbcaf2221925d61b018f252ca5b2113e8e/inc/breadcrumb.php#L1-L14

justinkruit:
The new commit f1d8939 checks if the child theme has the old function included, and if it does, it will use that old return value as the filter. This way the update won't be a breaking change (yet).

justinkruit:
@crftwrk How far do we want to go with adding filters?
For example, strings like this can also be wrapped in a filter to easily change without the need to copy the whole template file.
https://github.com/bootscore/bootscore/blob/00a418bbcaf2221925d61b018f252ca5b2113e8e/header.php#L55

crftwrk:
Think changing the classes of titles etc. will go a bit too much ;-)

justinkruit:
> Think changing the classes of titles etc. will go a bit too much ;-)

Didn't mean the classes in this case, but the text like "menu". Having such strings replaceable is what makes the filters in woocommerce so "powerful". We could also go that route.

crftwrk:
Understand,  I‘m in. We have to list all of them in the docs.

crftwrk:
> It could work like this (rough sketch)

Yes, this seems to be a good idea 🙏

page.php
```diff
@@ -16,14 +16,14 @@
...
-  <div id="content" class="site-content <?= bootscore_container_class(); ?> py-5 mt-5">
+  <div id="content" class="site-content <?= apply_filters('bootscore_container_class', 'container', 'page'); ?> py-5 mt-5">
     <div id="primary" class="content-area">
 
       <!-- Hook to add something nice -->
       <?php bs_after_primary(); ?>
 
       <div class="row">
-        <div class="<?= bootscore_main_col_class(); ?>">
+        <div class="<?= apply_filters('bootscore_main_col_class', 'col'); ?>">
```
crftwrk, line 26: This may create issues, because `col` isn't enough. If main content has for example a large image, `col` takes the entire width of the container and pushes sidebar down. It has to checked if a sidebar is active or not. See this very first PR https://github.com/bootscore/bootscore/pull/7.

crftwrk: Great 👍

zsvendo:
If possible, create the positioning of the sidebars or where they should appear. For example, natively switch sides or even not show it on the single-product page

justinkruit:
Time to rename all the filters, I want them with / instead of _ 😇

crftwrk:
Can we add filters to those margin and padding classes in templates?

crftwrk:
I mean those py-5 mt-5 classes next to the new container filter

https://github.com/bootscore/bootscore/blob/6651f8a0a53f36379cd121948a61212b2c2b3bfc/page.php#L19

justinkruit:
> I mean those py-5 mt-5 classes next to the new container filter
> 
> https://github.com/bootscore/bootscore/blob/6651f8a0a53f36379cd121948a61212b2c2b3bfc/page.php#L19

Done! https://github.com/bootscore/bootscore/pull/575/commits/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5

justinkruit:
@crftwrk got any more changes you'd like to see regarding filters?

crftwrk:
@justinkruit yes, possible to add filter to these classes? 

#### This will allow developers to change header and footer colors without touching the files:

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/header.php#L37

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/footer.php#L22

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/footer.php#L78

#### Filter to btn classes allows to quickly change buttons:

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/header.php#L91

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/template-parts/header/actions-woocommerce.php#L21

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/template-parts/header/actions-woocommerce.php#L31

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/template-parts/header/actions-woocommerce.php#L43

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/template-parts/header/actions-woocommerce.php#L49

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/template-parts/header/actions.php#L28

For classic widgets

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/searchform.php#L4

#### Regarding to https://github.com/bootscore/bootscore/pull/575#issuecomment-1816175137

Possible to do something with the sidebar in `woocommerce.php`?

https://github.com/bootscore/bootscore/blob/78b92fd662a02679ffb7008f2fb05ddf2d00e1b5/woocommerce.php#L34

justinkruit:
> @justinkruit yes, possible to add filter to these classes?

Damn that's gonna be a lot of filters 😂
I'll see what I can do and still keep the code clean, because that is my main concern.

crftwrk:
I thought about it again for a while and I think we should review and merge this asap and do changes in another PR. I have some ideas to simplify this up but this PR should be merged first. Think we should go this way, otherwise v6 will never come to an end.

https://github.com/bootscore/bootscore/pull/685 is important as well.

@justinkruit do you agree?

justinkruit:
> #### Regarding to [#575 (comment)](https://github.com/bootscore/bootscore/pull/575#issuecomment-1816175137)
> Possible to do something with the sidebar in `woocommerce.php`?

Just checked, and this should already be possible. The sidebar has a filter named `bootscore/sidebar/col_class`, and if you add a filter to that you should be able to determine using default WordPress functions on which page it is and change the classes.

justinkruit:
I think this version of the PR is ready to be reviewed and merged.

The buttons haven't been done, but for now that would lead to additional clutter in the theme. Maybe we could transform the button into a function to call with all the details in an array? Then the array can be added as a filter in the function itself.

Something like (first argument is just a name for the location and help the filter)
```php
<?= bootscore_button('offcanvas/search', ['text' => '<i class="fa-solid fa-magnifying-glass"></i><span class="visually-hidden-focusable">Search</span>', 'class' => 'btn btn-outline-secondary d-lg-none ms-1 ms-md-2 top-nav-search-md']);
```

crftwrk:
Tried 

```php
function container_class($string, $location) {
  if ($location == 'header') {
    return "container-fluid";
  }
  return $string;
}

add_filter('bootscore_container_class', 'container_class', 10, 2);
```

and

```php
add_filter('bootscore_load_fontawesome', '__return_false');
```

but nothing happens?

justinkruit:
Since the original commit, filters have been renamed. The ones you need are `bootscore/container_class` and `bootscore/load_fontawesome`

crftwrk:
Got it, works fine.

Possible to add multiple locations here?

```php
  if ($location == 'header') {
```

crftwrk:
No, I mean targeting multiple files in one function like if location header, footer, page-sidebar-left. If so, what is the snippet if i want to target containers only in header and footer?

justinkruit:
If you want to want to have a fluid container in both the header and footer, it will be like this:
```php
function container_class($string, $location) {
  if ($location == 'header' || $location == 'footer') {
    return "container-fluid";
  }
  return $string;
}

add_filter('bootscore/container_class', 'container_class', 10, 2);
```
As simple as that 😄

crftwrk:
I love it, but I need more information how to use them in the docs. Can you drop a quick example how to change roughly the main-col and sidebar width included the toggler?

justinkruit:
Writing out some examples of the filters, will drop them this evening!

crftwrk:
Thank you.

Should this function not wrapped in `if (!function_exists('bootscore_sidebar_col_class_sidebar')) {`?

https://github.com/bootscore/bootscore/blob/febfbabf0e69680ede381b012470c6641e93e00d/inc/columns.php#L21-L30

justinkruit:
> Should this function not wrapped in `if (!function_exists('bootscore_sidebar_col_class_sidebar')) {` to override in child?

In that case they can just add another filter 😄

justinkruit:
Here are a handful of examples that I could think of. A lot of things are possible with filters, and developers can make changes for niche situations in their own theme.

<details><summary>Disable FontAwesome</summary>

```php
add_filter('bootscore/load_fontawesome', '__return_false');
```

</details> 

<details><summary>Set a new logo</summary>

```php
function set_bootstrap_logo($logo, $size) {
  if ($size === 'small') {
    return get_stylesheet_directory_uri() . '/path/to/small-logo.svg';
  }

  return get_stylesheet_directory_uri() . '/path/to/normal-logo.svg';
}

add_filter('bootscore/logo', 'set_bootstrap_logo', 10, 2);
```

</details> 

<details><summary>Set the offcanvas title</summary>

```php
function set_offcanvas_title($title) {
  return 'My Menu';
}

add_filter('bootscore/offcanvas/navbar/title', 'set_offcanvas_title');
```

</details> 

<details><summary>Set the content spacer on an archive page of the post type "product"</summary>

```php
function set_content_spacer($spacer) {
  if (is_post_type_archive('product')) {
    return 'mt-5';
  }

  return $spacer;
}

add_filter('bootscore/content/spacer_class', 'set_content_spacer');
```

</details> 

<details><summary>Change the container class in the header and footer</summary>

```php
function container_class($string, $location) {
  if ($location == 'header' || $location == 'footer') {
    return "container-fluid";
  }
  return $string;
}

add_filter('bootscore/container_class', 'container_class', 10, 2);
```

</details> 

<details><summary>Make changes in inc/blocks/block-widget-archives.php</summary>

```php
function change_block_widget_archives($block_content, $block) {
  // do something with $block_content
  return $block_content;
}

add_filter('bootscore/block/archives/content', 'change_block_widget_archives', 10, 2);
```

</details> 

<details><summary>Disable the SCSS compiler</summary>

```php
add_filter('bootscore/scss/disable_compiler', '__return_true');
```

</details> 

<details><summary>Change the col size when the sidebar is active</summary>

```php
function change_content_col_size($col_size) {
  if (is_active_sidebar('sidebar-1')) {
    return "col-md-6 col-lg-7";
  }

  return $col_size;
}

add_filter('bootscore/main/col_class', 'change_content_col_size');

function change_sidebar_col_size($col_size) {
  return "col-md-6 col-lg-5 order-first order-md-last";
}

add_filter('bootscore/sidebar/col_class', 'change_sidebar_col_size');
```

</details> 

<details><summary>Remove order-first order-md-last from sidebar</summary>

```php
function remove_order_first_last($col_class) {
  return str_replace('order-first order-md-last', '', $col_class);
}

add_filter('bootscore/sidebar/col_class', 'remove_order_first_last');
```

</details>

crftwrk (approved):
Ok, got it and this is damn hot, thank you! As i see, there is no single recipe how to use them, but offers many possibilities. For the docs we should start with simple examples how to change the containers and main-col breakpoints and adding more examples from time to time.

crftwrk:
@justinkruit I am on the docs part for the filters right now, but I can't get this to work?

```php
function change_content_col_size($col_size) {
  if (is_active_sidebar('sidebar-1')) {
    return "col-md-6 col-lg-7";
  }

  return $col_size;
}

add_filter('bootscore/main/col_class', 'change_content_col_size');

function change_sidebar_col_size($col_size) {
  return "col-md-6 col-lg-5 order-first order-md-last";
}

add_filter('bootscore/sidebar/col_class', 'change_sidebar_col_size')

```

justinkruit:
Let me try that in a couple of hours when I'm home. It does seem correct.

crftwrk:
Thanks, it works for the sidebar but not for the main content?

justinkruit:
Where do you place the code to test?

crftwrk:
https://dev.bootscore.me/category/clerkship/ Sidebar is below the main-col

```php
function change_content_col_size($col_size) {
  if (is_active_sidebar('sidebar-1')) {
    return "col-lg-6 ";
  }

  return $col_size;
}

add_filter('bootscore/class/col/main', 'change_content_col_size');

function change_sidebar_col_size($col_size) {
  return "col-lg-6 order-first order-lg-last";
}

add_filter('bootscore/class/col/sidebar', 'change_sidebar_col_size');
```

justinkruit:
I mean, where in the code did you place it? File name and line

crftwrk:
Of course functions.php in child.

is it something here https://github.com/bootscore/bootscore/blob/main/inc/columns.php?

justinkruit:
> Of course functions.php in child.

So this is the reason. This happens because of the "priority" system in the filters. To quote the documentation:
> Used to specify the order in which the functions associated with a particular filter are executed.
Lower numbers correspond with earlier execution, and functions with the same priority are executed in the order in which they were added to the filter.

In this case, the child is loaded before the parent. So [this line](https://github.com/bootscore/bootscore/blob/36b96b04f00670c1bc5f9723aa3bec326c682a94/inc/columns.php#L30) is executed after the code in the child, overriding it. To make it work, add a priority of 11 to our new filter (10 being the default). Now it will be executed after the code of the main theme, effectively overriding it.

```php
function change_content_col_size($col_size) {
  if (is_active_sidebar('sidebar-1')) {
    return "col-lg-6 ";
  }

  return $col_size;
}

add_filter('bootscore/class/col/main', 'change_content_col_size', 11);
```

Fun fact: because the code in columns.php has already executed, the `$col_size` in the function above isn't `col` like by default, but already `col-lg-9`. This because it already got changed [here](https://github.com/bootscore/bootscore/blob/36b96b04f00670c1bc5f9723aa3bec326c682a94/inc/columns.php#L24).

crftwrk:
Got it, thanks. Sometimes things are so simple... ;-)

Issue #542, opened 2023-08-09
Cross Sell Products


gurpreetbhachu:
Hi,

Working on a practice project (eCommerce) on Version: 5.3.0

I have a few products under Cross Sell products for a product(Ex: item1).
So, the cross sell items are not displaying on the cart page after adding item1.

Am I missing something? or any settings?
Any suggestions?

Regards,
Gurpreet

crftwrk:
Hi Gurpeet,

you are right. Cross selling in cart is disabled by a function:

https://github.com/bootscore/bootscore/blob/3351cbe1ad9201d518b8432e283d02e54e9747a8/woocommerce/wc-functions.php#L110-L111

We are refactoring the entire WooCommerce part https://github.com/bootscore/bootscore/issues/333 since a while and the cross selling loop in cart do not fit yet. That's why cross selling is disabled for now, but it's planned to enable in v5.4.

If this is urgent for you, you can create a pull request and wrap the line in a pluggable function `if (!function_exists('function_name')) :`. Then you can simply add the empty function to your child and cross selling is enabled.

Hope that clears it up

Basti

PR #577, opened 2023-09-18
Enable cross sell products

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/542

### Test

- Go to shop https://dev.bootscore.me/shop/ and add "Album" and "Beanie with Logo" to the cart
- Go to cart page

cross sell

### DIFF
--- a/scss/bootscore_woocommerce/_wc_cart.scss
+++ b/scss/bootscore_woocommerce/_wc_cart.scss
@@ -136,6 +136,6 @@ WooCommerce Cart
   margin-bottom: $spacer;
 
   .col-md-6.col-lg-4.col-xxl-3 {
-    width: 50%
+    width: 50%;
   }
 }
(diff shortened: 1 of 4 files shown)


justinkruit:
Maybe there is a different way to do this, instead of overriding the col sizes with css which I personally don't find very clean.

Maaaayyybbee we can make the cols a filter? And write a check if it's the cross sells list, and then override the cols.

crftwrk:
> Maaaayyybbee we can make the cols a filter? And write a check if it's the cross sells list, and then override the cols.

Of course, any idea?

justinkruit:
@crftwrk I've added 2 commits to the PR, which changes the cols into a filter and makes it 6 cols wide on the cart page.

crftwrk:
Much more better, thank you!

justinkruit (approved):

Issue #589, opened 2023-09-30
[BUG] .top-button is clickable when not visible


JWestarp:
### Describe the bug

The scroll to top button is always clickable. 

Especially on the smartphone, it is easy to accidentally click on the invisible button and jump to the top of the page, which is very iiritating.

### Steps to reproduce

1. Move Mouse to right bottom corner and click.

PR #590, opened 2023-09-30
bug fix: #589 (added visibility tansition)

JWestarp:


crftwrk:
Thank you fore pointing this out!

Just 2 tiny things:

1 - We can skip `visibility: visible;` because Bootstrap has 

```css
.visible {
  visibility: visible !important;
}
```

in utilities.

2 - Using `.3s` is cleaner than `300ms` and used in the entire theme.


Overall code:

```scss
.top-button {
  right: 5%;
  bottom: 30px;
  opacity: 0;
  visibility: hidden;
  transition: visibility 0s linear .3s, opacity .3s;

  &.visible {
    opacity: 1;
    transition: visibility 0s linear 0s, opacity .3s;
  }
}
```

If you want to change this quickly, we are ready to merge.

JWestarp:
@crftwrk Thanks for the quick reply. Yes, i agree with you. 
I just changed that.

crftwrk (approved):
👍

PR #591, opened 2023-09-30
Smooth scroll to anchor and hide hash

JWestarp:
- `theme.js`: 
  - added function for smooth scroll to anchors
  - prevents click action
  - hides hash in url (for example when clicking .top-button
- `footer.php` 
  - added existing anchor target for .top-button instead of '#'

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -84,7 +84,7 @@
 </footer>
 
 <!-- To top button -->
-<a href="#" class="btn btn-primary shadow top-button position-fixed zi-1020"><i class="fa-solid fa-chevron-up"></i><span class="visually-hidden-focusable">To top</span></a>
+<a href="#page" class="btn btn-primary shadow top-button position-fixed zi-1020"><i class="fa-solid fa-chevron-up"></i><span class="visually-hidden-focusable">To top</span></a>
 
 </div><!-- #page -->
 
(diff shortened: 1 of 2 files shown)


crftwrk:
Thank you for this PR.

We had a jQuery script like this before, but were very happy to remove https://github.com/bootscore/bootscore/pull/114 it since `scroll-behaviour: smooth` is part of Bootstrap and supported by all major browers https://caniuse.com/css-scroll-behavior.

I agree that removing the hash in URL would be nice, but I do not think that we switch back to jQuery smooth scroll version.

Set this to draft to avoid accidentally merging and let's wait what other users think about this.

justinkruit:
Personally I don't think removing the hash from the URL should be default in the theme. It's a nice addition for when it's needed, but I feel like this type of behavior is more up to the developer instead of having it by default.

JWestarp:
@crftwrk 
Interesting! I didn't know about `scroll-behaviour: smooth`. At least it doesn't work with my Firefox 119.0b3 (64-Bit) on macOS 12.6.7. That's why I came up with the idea to solve this myself.
Then I'll see what's wrong with my browser.  :)

justinkruit:
Do you have animations turned off on your device? Firefox and Chrome take over some settings from the OS to change their behaviour. This includes smooth scrolling.

JWestarp:
@justinkruit Thank you very much! That was the problem. In my Firefox 'Enable smooth scrolling' was not activated.

I think i can't let the user choose how the websites of our customers look or feel like. so i think a java-script solution is better to keep the quality the same across all users.
even i am too stupid to set it initially the way it would feel better.

thank you all for your input!

PR #602, opened 2023-10-14
Improve header user, cart and checkout

crftwrk:
… mini-cart in checkout with back-to-cart button


justinkruit (approved):
Looks good to me! We could even expand this behavior in #575 and add filters to the ifs to always show them without the need to change the template part.

crftwrk:
> We could even expand this behavior in https://github.com/bootscore/bootscore/pull/575 and add filters to the ifs to always show them without the need to change the template part.

Yes, I'm in 👍

PR #616, opened 2023-10-30
Replace sr-only with visually-hidden

### LINKED ISSUE #614: Pagination loads sr-only span, which is deprecated in BS 5.3
### Describe the bug

It appears to me the parent theme loads `Page navigation` span with class `sr-only` which is not anymore part of Bootstrap, instead, `visually-hidden` should be used, if I understand https://getbootstrap.com/docs/5.3/getting-started/accessibility/#visually-hidden-content properly. And as a matter of facts I can see (and I am not señor yet) the `Page navigation` string on the screen even if I am not using assistive tech. 

It does not appear as an issue as long you keep the FA styles - but as soon you dequeue them (which I do regularly, as I use BS Icons), the `sr-only` has no CSS anymore.
That CSS indeed comes from FA styles as far I can see.

My suggestion is to fully replace the class with `visually-hidden` so even if one dequeues FA, the assistive class does not "fail"

I might miss something.. please let me know before I do a PR

### Steps to reproduce

- Install Theme 
- Dequeue FA
- Load an archive that has more than one page.

### Screenshots and Additional Info

Screenshot 2023-10-29 at 15 42 39

https://github.com/bootscore/bootscore/blob/042759bc6f7843e1ba1d39ead8a80fb45f5c75b2/inc/pagination.php#L36
https://github.com/bootscore/bootscore/blob/042759bc6f7843e1ba1d39ead8a80fb45f5c75b2/inc/pagination.php#L49

### Website link

_No response_

smileBeda:
- [Fixed] Replace `sr-only` with `visually-hidden` as per Bootstrap 5.3. Fixes #614

### DIFF
--- a/inc/pagination.php
+++ b/inc/pagination.php
@@ -33,7 +33,7 @@ if (!function_exists('bootscore_pagination')) :
 
     if (1 != $pages) {
       echo '<nav aria-label="Page navigation" role="navigation">';
-      echo '<span class="sr-only">' . esc_html__('Page navigation', 'bootscore') . '</span>';
+      echo '<span class="visually-hidden">' . esc_html__('Page navigation', 'bootscore') . '</span>';
       echo '<ul class="pagination justify-content-center mb-4">';
 
       if ($paged > 2 && $paged > $range + 1 && $showitems < $pages) {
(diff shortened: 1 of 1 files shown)


crftwrk (approved):
👍🙏

Issue #619, opened 2023-11-06
HTML5 compliance for pagination


smileBeda:
### Describe the bug

As per HTML5 the `nav` element does not require a `navigation` role, we can remove it.

Not sure if you want to keep it though. Pre-PR clarifying that.

### Steps to reproduce

1. Install Bootscore theme
2. Run HTML5 validation on a paginated element

### Screenshots and Additional Info

Source at https://github.com/bootscore/bootscore/blob/cc2f622eea250a683efaef29e47fc6ac22d2c6fe/inc/pagination.php#L35
Suggested edit: remove `role="navigation`

crftwrk:
Agree, PR?

Issue #617, opened 2023-11-03
Update all occurrences of wp_localize_script to wp_add_inline_script


smileBeda:
### Describe the bug

The problem is `wp_localize_script` - while it has been the de-facto go to to add inline scripts for years - is the wrong tool to use for this task. And it presents security issues.

1. `wp_localize_script` is strictly intended to _localise_ data
2. `wp_localize_script` is not filtered, thus, it is impossible to add CSP nonces to its script tag, effectively making it impossible implement strict CSP (and herein lies the security issue)

The solution is to use `wp_add_inline_script` instead. This appears to be filtered and the CSP can be added with something like this:
```
add_action('send_headers', 'set_csp_header');
function set_csp_header() {
    $csp_nonce = base64_encode(random_bytes(16));
    wp_cache_set('csp_nonce', $csp_nonce);
    header("Content-Security-Policy: script-src 'self' 'nonce-{$csp_nonce}'");
}

add_filter('wp_inline_script_attributes', 'add_nonce_to_inline_script', 10);
function add_nonce_to_inline_script($attr) {
        $csp_nonce = wp_cache_get('csp_nonce');
	if ( empty( $attr ) ) {
		return $attr;
	}
	return array_merge(
		$attr,
		array(
			'nonce' => $csp_nonce,
		)
	);
    
}
````

This then allows to have a `nonce` in the inline script tag that allows strict CSP

@crftwrk - do you have concerns about backwards compatibility? The `wp_inline_script_attributes` hook unfortunately does not exist before WP 5.7
But it is the right and only way going forward. For me personally this is a little showstopper since right now, after developing my new site on a staging (where obviously I did neither consider nor implement CSP) I did not notice at all that the older wp_localize_script is used by Bootscore. As I went live and added my plugin which handles my CSP... I saw the error logs growing in the console of course :P
As of now there is no other solution than editing the parent theme...

Let me know if you are fine with requiring WP 5.7 and I'll do some PRs

### Steps to reproduce

1. Install Bootscore Theme
2. Implement strict CSP 
3. Notice JS errors in console

smileBeda:
It is the IE script, yes. I am fine with dropping it altogether, if you agree.

crftwrk:
Thought that `wp_localize_script` is used for offcanvas cart translation as well. But I'm wrong, it's only used in IE alert. From my site, this two snippets can be deleted:

[...]

PR #620, opened 2023-11-08
HTML5 Compliance, Remove IE warning alert

smileBeda:
- [Changed] Removed navigation role from nav element as per HTML5 standards. See #619 
- [Removed] Remove IE Support and warning. See #617 

Closes https://github.com/bootscore/bootscore/issues/619
Closes https://github.com/bootscore/bootscore/issues/617


crftwrk (approved):
LGTM 👍

@justinkruit if you agree, please hit the merge button.

justinkruit (approved):
All looks good, happy we're removing the IE warning. If you still use that, no way to change their mind anyways 😂

Issue #419, opened 2023-03-03
Some Irritations concerning the favicon docu:


kreincke:
Due to the fact, that you ["seriously recommend replacing ALL icons and colors before going live to show site correct in all browsers and search results"](https://bootscore.me/documentation/header-php/), I've just tested the way you've described. At least for the favicons, there are some minor issues.

Basis: First I've created the favicon zip file as you've described it in your docu concerning the header.php. Then I've uploaded the set of replaced facicon-files. At last I've checked the results via Favicon checker:

* iOS Safari: works.
* Android Chrome: works.
* Windows 8 and 10: Feedback "There is no Browser Configuration file at https://xyz/browserconfig.xml ". If I put it manually in the ROOT directory xyz, the feedback is: "The Browser Configuration file is present. [But] https://fxyz/mstile-150x150.png is missed." That's valid, because the file is still inside the bootscore-child/img/favicon. If I look at the header section of head.php, I cannot find any entry containing "echo get_stylesheet_directory_uri(); ?>/img/favicon/browserconfig.xml". So, I assume, that anything mishappens with the integration of that file or the description in your docu has forgotten to mention, that both, the file browserconfig.xml and the file mstile-150x150.png have to be put into the ROOT manually.
* Mac OS X El Capitan Safari: works
* Classic, desktop browsers: works. But the test remarks, that favicon is ill-formed and does not contain a layer of 48x48. (But the file can be read by GIMP and contains all 3 Layers) This is a bit silly, because the Favicon checker is a tool of https://realfavicongenerator.net/. So, I think, we can ignore it: the tool fails its own test.

Additionally, there are some irritations in the steps you described:

* YOU: "Adjust colors for Safari pinned tab icon and Windows phones in lines 26, and 27." WE: In which file shall we adjust the values? Really in safari-pinned-tab.svg? Line 26 and 27 seem not to contain general color information. 
* YOU: "Adjust theme-color in line 28, this should be your primary-color." WE: In which file shall we do that.

Perhaps you could improve the descriptions and fix the browserconfig.xml issue - if and when you have some time in between

Many thanks for your patience. KR

crftwrk:
The favicons are described in the docs for header.php https://bootscore.me/documentation/header-php/. That‘s why you will find thise lines in header.php.

crftwrk:
It's not possible to edit docs via PR, because this are WordPress posts and not hosted here. You can drop text here and we will publish it on bootscore.me.

[...]

PR #634, opened 2023-11-22
Remove favicons and links in `header.php`

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/419
PR in child https://github.com/bootscore/bootscore-child/pull/34

This PR removes the favicons and links to them in `header.php`

### Why?

- It's against the WordPress codex. These links were one (now the last) reason why we couldn't publish to the WP theme repository.
- Buggy with Android Chrome and Windows 8/10.
- Users do not take this step serious enough. Most developers works on Windows with Chrome, and didn’t check on any other device. I'm an Apple user and I see beautiful designed sites but viewing them in Safari shows still the blue b :-/

Think we should give this back to the Customizer and mention in the docs that the 90kb plugin https://wordpress.org/plugins/favicon-by-realfavicongenerator/ leads to the best results.

When merged, we can theoretically ask for a theme review in the WP repository, question is if we want to go this way.

@justinkruit if you agree, please merge.


justinkruit:
> I'm an Apple user and I see beautiful designed sites but viewing them in Safari shows still the blue b :-/

That's actually quite funny 😂 

> When merged, we can theoretically ask for a theme review in the WP repository, question is if we want to go this way.

Personally, not the biggest fan. I think this theme relies on a developer making changes to make it fit their needs. For me the WP repository is more suited for non-developers.

justinkruit (approved):

crftwrk:
> For me the WP repository is more suited for non-developers.

Agree

PR #635, opened 2023-11-23
Replace `my-account-offcanvas.php` with `my-account.php` in user offcanvas

crftwrk:
> [!IMPORTANT] 
> Todo: Hide user and offcanvas in checkout page - duplicate id's.

Thanks to https://github.com/bootscore/bootscore/pull/602 there is no conflict with double id's anymore and we can include `my-account.php` instead the own file `my-account-offcanvas.php` in user offcanvas. This will further reduce maintenance effort.

Tested on several live shops and works fine. Is already live on bootscore.me as well.

| Before  | After |
| log-in-before  | log-in-after  |
| logged-in-before  | logged-in-after  |


scss/bootscore_woocommerce/_wc_my_account_offcanvas.scss
```diff
@@ -2,25 +2,34 @@
...
     }
   }
-}
\ No newline at end of file
+
+  // my-account
+  .col-md-4,
+  .col-md-8 {
+    width: 100%;
+  }
```
justinkruit, line 30: Do we want to achieve this using filters instead? Just like we did with https://github.com/bootscore/bootscore/pull/577/commits/696e5024d7f776f269796e961ec23cfa2be457f6? Not really sure how we filter if it's the offcanvas, but I'm sure we can find a way. But I'll also be fine with this solution hahaha.

justinkruit (changes requested):

scss/bootscore_woocommerce/_wc_my_account_offcanvas.scss
crftwrk: The thing is that we have to readd the WC template to the file and I'm very happy that this file is removed.

justinkruit: But the cols are set somewhere, right? We can add the filters in there.
But looking more closely, it being loaded from shortcode will complicate the way to find out where the filter is coming from. I think this is the easier method then.

crftwrk: Ups, sorry. My bad. Here are the files with the cols:

- https://github.com/bootscore/bootscore/blob/include-my-account.php-in-user-offcanvas/woocommerce/myaccount/navigation.php
- https://github.com/bootscore/bootscore/blob/include-my-account.php-in-user-offcanvas/woocommerce/myaccount/my-account.php

Filter will be much more better

justinkruit: I still haven't really thought of a way to implement the filter here, as I can't really do a check if the current location is the offcanvas. Will need to think about it more, or we do this solution for now.

crftwrk: Think the CSS solution is fine, because the columns are in a Woo template and if they do an update to this file, I will try to find a way to remove them. But I'm sure Woo will do something completely new with blocks there in future. So I think we not should pay too much attention to that.

As mentioned, this PR will have a conflict with https://github.com/bootscore/bootscore/pull/647 which is much more urgent. I will fix this after https://github.com/bootscore/bootscore/pull/647 is merged.

justinkruit:
What's the reason changes have been made on an old PR? That doesn't seem right

crftwrk:
> What's the reason changes have been made on an old PR? That doesn't seem right

Because this is an important PR and had conflicts with a previous one. I will make this ready after WooCommerce 8.6 release (today).

Issue #640, opened 2023-11-27
Rewrite main.scss


justinkruit:
The main.scss can be expanded so people have more control over Bootstrap. This can be achieved by not loading Bootstrap in one go, but in parts. An example of that can be found [here (option B)](https://getbootstrap.com/docs/5.3/customize/sass/#importing).

This will improve #306, and make things mentioned in https://github.com/orgs/bootscore/discussions/639 possible.

PR #645, opened 2023-12-01
Rewrite main.scss

justinkruit:
In issue #640 we highlighted that a rewrite of the `main.scss` is needed to give the users more control over Bootstrap. With this PR that will be achieved.

The variables are set before Bootstrap so everything is better overridable. 
A new `_bscore_maps.scss` has been added so maps can also be overridden. 
And lastly a new `_bscore_utilities.scss` has been added and loaded _after_ the Bootstrap utilities so they can be added upon ([docs](https://getbootstrap.com/docs/5.3/utilities/api/#using-the-api)).

I've tested the compiling locally, and nothing seems to be missing.

@crftwrk If you like it, merge it. 😜

### DIFF
--- a/scss/_bscore_utilities.scss
+++ b/scss/_bscore_utilities.scss
@@ -1,0 +1,4 @@
+/*
+  Use this file to add your own custom utility classes.
+  https://getbootstrap.com/docs/5.3/utilities/api/
+*/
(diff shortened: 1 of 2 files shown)


crftwrk:
That's cool, thank you! Just two things:

- `@import "bootstrap/helpers";` is missing. Check the to-top button, `visually-hidden` elements are visible.
- We need this in child as well.

crftwrk (approved):
Looks good! Waiting for the child PR.

PR #647, opened 2023-12-04
Reduce compiled css file size

crftwrk:
Had a look to the compiled CSS and file size is big, too big. It's overall 455kb which includes 200kb Bootstrap, 30kb theme and the rest is the WC part. This PR reduces compiled minified CSS to 275kb which is absolute ok compared to other themes.

Figured out that using `@extend` on forms and buttons is a bad idea, because they have all this disabled, focus, focus-within, isvalid, is-invalid etc. states. This states creates in combination with the very bad WC classes huge selector strings which are only visible if you check one of this states. We do not need this states here.

Where possible, I readded the filters, but they do not catch all forms and buttons. For the rest i created custom form, custom check/radio with only :focus state and custom buttons with only :hover state and assigned them to the WC classes. This custom components are built with Bootstrap variables. Values can be changed as usual in `_bscore_variables.scss`.

The SCSS looks ugly but works and reduces 180kb file size. Tested in on several live shops and works fine, already live on bootscore.me as well. 

Test: Check forms, checks and radios in cart, checkout and dashboard. Not each file must be reviewed, viewing the frontend is enough.

PR will have a conflict with https://github.com/bootscore/bootscore/pull/635, but I can fix this later. This one here seems to more important for me.

@justinkruit think this is an serious issue, IYLIMI ;-)

ps. Using `@extend` on other elements like badges or alerts have no negative effect.


justinkruit (changes requested):
The colors of the Coupon input don't seem to match, and the box shadow misses the animation.
| Current | Docs |
|--------|--------|
| image | image |

Other than that I can't spot anything.

crftwrk:
Thank you for pointing this out, it has been fixed https://github.com/bootscore/bootscore/pull/647/commits/a68442ad96bd96e1931ce28d07868497f02e3685. Please check the cart coupon input https://bootscore.me/cart/.

justinkruit (changes requested):
image

Just spotted that the disabled outline button is also different, WC is setting a background color. Tbh not really intrusive, unless we want to force the Bootstrap style on it completely. Up to you if we also want to change that.

crftwrk:
Fixed https://github.com/bootscore/bootscore/pull/647/commits/40761c461b014ffdc498260c244b90344496517a

justinkruit:
> It's overall 455kb which includes 200kb Bootstrap, 30kb theme and the rest is the WC part. This PR reduces compiled minified CSS to 275kb which is absolute ok compared to other themes.

Is this comparison with expanded style, or with compressed? Just wondering. Reference: https://github.com/bootscore/bootscore/blob/cfbc06855668c33bbe3a19b6d3606106a4943c52/inc/scss-compiler.php#L51-L61

crftwrk:
It's the compressed one.


| Before  | After |
| before  | after  |

Pushed 2 more commits, one fixes the coupon bg in dark mode and the other one replaces the product comment jQuery checkbox with the new SCSS. I found no more bugs, seems to work well...

justinkruit (approved):
All seems well.

PR #660, opened 2023-12-20
Remove filter to enable classic widgets by default

crftwrk:
This PR removes the filter which enables classic widgets by default. I think we should let users decide by them self if they want to use Gutenberg or classic widgets by adding the filter to their child or installing the https://wordpress.org/plugins/classic-widgets/ plugin. 

The reason why I've added the filters was the search widget. Now Gutenberg search widget is used which is very ugly, but maybe this forces us to do something more to support blocks. When PR is merged, I will fix the Gutenberg search block.

@justinkruit IYLIMI!

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -6,7 +6,7 @@
  * @link https://developer.wordpress.org/themes/basics/theme-functions/
  *
  * @package Bootscore
- * @version 5.3.3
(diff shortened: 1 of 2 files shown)


justinkruit (approved):
Seems fine to me!

Issue #471, opened 2023-05-04
[Feature Request] Add BS5 Navwalker support for WP NAV Widgets


seanvandermolen:
### Describe the feature that you'd like

For consistency and proper classes NAV widgets should use the BS-5 Navwalker by default.

The function is simple enough:

```
add_filter('widget_nav_menu_args', 'my_args'); //for menus in widgets
function my_args($args) { //$args is only argument needed to add custom walker
   return array_merge( $args, array(
        'walker' => new bootstrap_5_wp_nav_menu_walker(),
   ) );
}
```

The Navwalker would need to be updated to remove 'nav-link' discovered in  LI  elements.

Add after line 58 in class-bootstrap-5-navwalker.php
$class_names = str_replace('nav-link ', '', $class_names);

### Why is this feature needed? What problem does it solve?

Any NAV menu will require BS classes which the BS-5-Navwalker solves for the most part.

### Checklist

- [X] I have searched the issues and this feature has not been requested.

crftwrk:
Thank you for this feature request. I've been thinking about this for a while and generally it's a good idea. But there's a reason why we shouldn't do that, let me explain.

Theme disabled Gutenberg widgets and enabled classic widgets by default. In order to keep up with the times, we have now reactivated Gutenberg in widgets https://github.com/bootscore/bootscore/pull/660 and we will add Bootstrap support for them step-by-step.

By enabling GB in widgets, existing nav-widgets move into legacy mode and support for them in WordPress will end. There is currently no GB nav-widget available, but if WP adds a new nav-widget block that will not be supported by the nav-walker. If WP adds a new nav-widget block, we can do nav support for them by adding a filter.

PR #662, opened 2023-12-24
Add support to Gutenberg block widgets

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/471

Because https://github.com/bootscore/bootscore/pull/660 disables classic widgets we should provide support for block widgets. This PR adds Bootstrap to following block widgets:

- Search `form-control/btn-outline-secondary/input-group/btn text/icon`
- Categories `select/list-group`
- Archives `select/list-group`
- Latest Posts `list-group`
- Latest Comments `list-group`
- Calendar `table`

Demo: https://dev.bootscore.me/theme/blocks/

In the sidebar are the widgets that come with a new WP installation wrapped in `.card`, in main content are the same widgets with different/advanced settings.

I have stolen this idea from the Understrap theme using `render_block_core` for this purpose and adapted it to our needs.

### Cards in sidebar

To get widgets in sidebar work properly, `.card` component has been removed in widget register. Cards can be added by using Group blocks and adding classes to the blocks advanced settings:

```html
Group class card
| - H2 class card-header h6
| - Group class card-body
  | - Block widget
```

```html
Group class card
| - H2 class card-header h6
| - Block widget Categories/Archives/... class list-group-flush
```

### Search form

Replacing the WP search icon with Font Awesome works only if svg is formatted this way. Overriding function in child must have same formatting, including the indentation tab for the function. Don't ask me why.

```php
/**
 * Change search btn and icon
 */
  function bootscore_block_widget_search_classes($block_content, $block) {

    $search  = array(
      'wp-block-search__input ',
      'wp-block-search__input"',
      'wp-block-search__button ',
      '<svg class="search-icon" viewBox="0 0 24 24" width="24" height="24">
                    <path d="M13 5c-3.3 0-6 2.7-6 6 0 1.4.5 2.7 1.3 3.7l-3.8 3.8 1.1 1.1 3.8-3.8c1 .8 2.3 1.3 3.7 1.3 3.3 0 6-2.7 6-6S16.3 5 13 5zm0 10.5c-2.5 0-4.5-2-4.5-4.5s2-4.5 4.5-4.5 4.5 2 4.5 4.5-2 4.5-4.5 4.5z"></path>
                </svg>'
    );
    $replace = array(
      'wp-block-search__input form-control ',
      'wp-block-search__input form-control"',
      'wp-block-search__button btn btn-primary ',
      '<i class="fa-solid fa-glasses"></i>'
    );

    if (isset($block['attrs']['buttonPosition']) && 'button-inside' === $block['attrs']['buttonPosition']) {
      $search[]  = 'wp-block-search__inside-wrapper';
      $replace[] = 'wp-block-search input-group';
    }

    return str_replace($search, $replace, $block_content);
  }
```

### What else?

- Moved outdated scripts & styles to deprecated
- Renamed `_recycle_bin.scss` into `_deprecated.scss`
- Renamed `_wc_recycle_bin.scss` into `_wc_deprecated.scss`

@justinkruit IYLIMI


justinkruit:
@crftwrk you still got more changes to do? 😉

crftwrk:
@justinkruit not yet, it‘s good for now.

inc/blocks/block-widget-categories.php
```diff
@@ -0,0 +1,53 @@
...
+        'cat-item list-group-item list-group-item-action d-flex justify-content-between align-items-center',
+        'current-cat active',
+        '<a class="stretched-link text-decoration-none"',
+        '<span class="badge bg-primary-subtle text-primary-emphasis">',
+        '</span>'
+      );
+
+      $block_content = str_replace($search, $replace, $block_content);
+    }
+
```
justinkruit, line 49: Here we can also add our own filters potentially, so people can change the content after our changes (could be done in #575). Either that or they can hook after the `render_block_core/categories` hook. What do you say?

justinkruit (commented):
Only got the one comment, other than that; impressive!

inc/blocks/block-widget-categories.php
crftwrk: Yes, good idea! 

Think those filters are a huge step and I can't wait for https://github.com/bootscore/bootscore/pull/575. BTW, we should add a filter for those `py-5 mt-5` classes as well:

```php
<div id="content" class="site-content <?= apply_filters('bootscore_container_class', 'container', 'page'); ?> py-5 mt-5">
```

Think this is very useful if users removes the `fixed-top` in `header.php`.

justinkruit: My idea is to merge this PR, and work on adding more filters (also for these changes) in #575. Agree?

crftwrk: Full agree👍

justinkruit (approved):

Issue #661, opened 2023-12-24
[BUG] Woocommerce Products Shortcode


andreiflorentin:
### Describe the bug

Hello,
When I use shortcodes to display products not works pagination. For example with this [products columns="4" on_sale="true" per_page="28" paginate="true" ] not display pagination. Do you can help me, please?

crftwrk:
Hi,

thank you for pointing this out, there is a linked PR https://github.com/bootscore/bootscore/pull/663 which fixes that.

Until this PR is merged, go to your main-theme's folder `woocommerce/loop/pagination.php`and replace the entire content with tis one https://github.com/bootscore/bootscore/blob/Fix-WC-pagination/woocommerce/loop/pagination.php.

BTW, `columns="4"` has no effect because columns are overridden by the theme. Using `[products on_sale="true" per_page="28" paginate="true"]` is enough.

andreiflorentin:
It's ok now. Thank you!

crftwrk:
You‘re welcome. I reopen this issue because there is a linked PR which closes this when merged.

PR #663, opened 2023-12-27
Fix WC pagination.php

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/661

Demo https://dev.bootscore.me/shop/wc-shortcodes/

This PR fixes the WC pagination if a shortcode has limited products and pagination is set to true `[products on_sale="true" per_page="2" paginate="true"]`.

@justinkruit IYLIMI

### DIFF
--- a/woocommerce/loop/pagination.php
+++ b/woocommerce/loop/pagination.php
@@ -1,5 +1,4 @@
 <?php
-
 /**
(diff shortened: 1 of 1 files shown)


justinkruit (approved):
Looks good to me! 👍

PR #95, opened 2022-02-10
add laravel mix instance phpsass compile and new things

vahidalvandi:
hi.
Based on our experiences working with Laravel, I have used them in other WordPress projects, and here I have added them as very practical and attractive ideas to this project. I hope you will approve it.

**Description of changes that have occurred**

1. added [](https://github.com/laravel/laravel/blob/9.x/package.json) in root project and run npm i command to install dependancyes 
2. disable  require_once 'inc/scss-compiler.php';   and   bootscore_compile_scss();   in functions.php
3. rename **bootscore/css/scss/bootstrap.min.scss** to **bootscore/css/scss/style.scss**
4. remove complete bootstrap/scss/bootsrap Because we only need mixers here and importing all styles will slow down the scss compilation process and also create duplicate code in srtyle.css, bootsrap.min.css
5. And finally, because of your need for alert.scss to bootrasp utility completely, Kant, of course, we have to talk about this

Issue #94, opened 2022-02-10
laravel mix 


vahidalvandi:
I think laravel mix should be used to manage sass
i am laravel developer and have same idea for incrase sass compile and incrase sass build and watch time during development wp templete .

justinkruit:
The idea behind behind bootScore is that it's easy to add to an existing Wordpress installation, on any PHP environment. If we would use anything like Laravel, the setup would be incredibly more difficult, and not what bootScore is intended to be.

vahidalvandi:
We don't need install laravel .

Laravel ream created a Web pack module as name laravel mix

https://laravel-mix.com/

This just need npm

vahidalvandi:
Dear npm is very modern from your sass compilers. Worldwide frontend
developers work with npm also is very fast and after development ended with
command
Npm run production create final style.css

justinkruit:
I'm very familiar with NPM, I'm a developer myself. Still, it's simply not how bootScore is intended. No additional tools are needed to make bootScore work as a theme, while with NPM the user is required to know how a terminal works, and have access to one on their simple webhosting (almost none do).

justinkruit:
> Agree. Name bootstrap.min.css and folder structure...

Then my vote definitely goes for the first branch.

> Tried already, but failed. Think a source map is very useful for debugging.

Alright, I'll probably give it a shot this weekend! I'll create a new issue for it

PR #667, opened 2024-01-06
ViteJs integration

mahmoud-alawad:
Add viteJs support


crftwrk:
Hi Mahmoud,

thank you for this PR. Integrating Vite means that users have to learn Vite and installing something before theme is ready to use. This makes it useless for 99% of our users. Bootscore has everything installed locally and uses scssphp to compile SCSS without doing anything in the code. So, we keep everything as it is and do things as simple as possible

See as a reference:

- https://github.com/bootscore/bootscore/issues/94
- https://github.com/bootscore/bootscore/pull/95

PR #670, opened 2024-01-13
Add support to WC categories block widgets

crftwrk:
Similar to https://github.com/bootscore/bootscore/pull/662, this PR adds support to Woo categories list/select widget block.

See https://dev.bootscore.me/theme/blocks/. The widgets are at the end of the sidebar and at the end of the main content are the same widgets with different settings.

Think we should merge this and add filters in https://github.com/bootscore/bootscore/pull/575. After https://github.com/bootscore/bootscore/pull/575 is merged we should update the libraries, do a PR with changing the logo and go for a new release, agree?

@justinkruit IYLIMI!


justinkruit (changes requested):
In the product-categories I'm missing a hover bg color on the button.

> Think we should merge this and add filters in https://github.com/bootscore/bootscore/pull/575. After https://github.com/bootscore/bootscore/pull/575 is merged we should update the libraries, do a PR with changing the logo and go for a new release, agree?

Yes I agree. Planning this for February?

crftwrk:
You're right. `input-group-text` has no `hover` state. Changed the button to `btn-outline-secondary` similar to the search form. Check https://dev.bootscore.me/theme/blocks/

> Yes I agree. Planning this for February?

Perfect

justinkruit (approved):
All seems good

Issue #631, opened 2023-11-21
[Feature Request] Quantity Swatches for sidebar cart (no reload)


username0136:
### Describe the feature that you'd like

I'm requesting the addition of quantity swatches for the sidebar cart. [Here's a visual representation](https://github.com/bootscore/bootscore/assets/102938585/c3be87ea-03f6-45f4-ab4d-46ccaa1048d9), and they should operate seamlessly with Ajax.

![image](https://github.com/bootscore/bootscore/assets/102938585/c3be87ea-03f6-45f4-ab4d-46ccaa1048d9)

For implementation reference, check out the same feature in the [selleradise-lite theme repository](https://github.com/redoxbird/selleradise-lite/). This functionality is already integrated into that theme.

### Why is this feature needed? What problem does it solve?

This enhancement significantly improves the customer experience.
Currently, it's cumbersome for customers to navigate to the cart page, modify quantities, and then click the update cart button. With quantity swatches in the sidebar cart, this process becomes more **straightforward**.

I usually hide the "view cart" button to streamline the shopping experience, encouraging users to utilize the sidebar cart. This not only simplifies their journey but also allows for a quicker checkout.

### Checklist

- [X] I have searched the issues and this feature has not been requested.

crftwrk:
Totally agree that this will be a nice feature. However, this seems to be a tough job and the question is: Are you able to contribute and work on this?

Quickly checked the selleradise theme and it seems that it does not use the `mini-cart.php` for the sidebar cart. It has to deeply figured out how the cart works there.

Another option is to modify the mini-cart with quantity https://gist.github.com/ashokmhrj/fda5a23f7bdfc55749bd2bd53c64a8ca.

Third option is trying to hard code the new mini cart block in the theme https://developer.woo.com/2023/06/16/best-practices-for-the-use-of-the-cart-fragments-api/. This block has already the qty feature. Since Woo started to replace the classic templates with blocks, this seems to be the most future proved option.

Let me know what you think.

PR #675, opened 2024-01-30
Add theme support for block template-parts

crftwrk:
This PR allows to use blocks as template parts. In Backend Appearance > Template Parts can now the mini-cart edited in more detail and we can hard code it in the theme like https://github.com/woocommerce/storefront/pull/2100.

Think in the long-term, our offcanvas-cart should be replaced by the mini-cart block https://github.com/bootscore/bootscore/issues/631.

Did roughly some changes to demonstrate how it works:

- Go to https://dev.bootscore.me/shop/ and add something to the cart
- Scroll down to the footer, there you will find the Woo mini-cart block. Open cart and check buttons, they have classes `btn btn-primary` etc.

@justinkruit IYLIMI

### DIFF
--- a/inc/theme-setup.php
+++ b/inc/theme-setup.php
@@ -4,7 +4,7 @@
  * Theme setup
  *
  * @package Bootscore 
- * @version 5.3.4
+ * @version 5.4.0
  */
 
 
@@ -32,7 +32,9 @@ if (!function_exists('bootscore_setup')) :
     */
     load_theme_textdomain('bootscore', get_template_directory() . '/languages');
 
-    // Add default posts and comments RSS feed links to head.
+    /*
+     * Add default posts and comments RSS feed links to head.
+    */
     add_theme_support('automatic-feed-links');
 
     /*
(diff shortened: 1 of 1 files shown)


justinkruit:
Personally, not the biggest fan of the block changes. This because it feels like giving over control to WooCommerce which has a bigger tendency to break stuff (see the alerts recently).

I'm also just seeing the block template parts system for the first time, and not entirely convinced. Will we be taking steps back from coding? Because Gutenberg has limited coding options in our current setup, which we can't really change unless we want to change how the theme works (making blocks with NPM compiling).

crftwrk:
> Will we be taking steps back from coding? Because Gutenberg has limited coding options in our current setup, which we can't really change unless we want to change how the theme works (making blocks with NPM compiling).

No, I do not mean to create a block theme, I like it as it is. But I think that the entire WP goes more and more to blocks and we have to find a way to be prepared. If for example Woo will remove the cart fragments entirely in future, we need an alternative for our mini-cart. This PR gives us the option to play around with it and maybe the mini-cart block can replace our one by hard coding it to the theme. This PR just allow us to edit blocks in more details and use them as template-parts, nothing more.

justinkruit:
Sorry for taking so long, been busy.

Alright, I agree we can and should look into this. But do we want to enable this for this upcoming release? Or wait to expand on this on the release after that?

crftwrk:
This has no effect to the theme but gives us more freedom to play around with blocks. Should be merged as soon as possible.

PR #678, opened 2024-02-07
Revert "Fix wc 8.5 alerts"

crftwrk:
Reverts bootscore/bootscore#671

WC 8.6 fixes the issue with the broken alerts https://developer.woo.com/2024/02/06/woocommerce-8-6-rc-classic-notice-templates-for-classic-themes-more/. In this case we should revert bootscore/bootscore#671 and using our previous alerts again, because they are proved and no WC template overrides for them are used in theme. 

#### Test

- WooCommerce 8.6-RC.1 is installed.
- https://dev.bootscore.me/product/beanie/. Add product twice to the cart to see success and danger message, check alert classes. Classes `woocommerce-message`, `woocommerce-error` and `woocommerce-info` for the empty cart are used again.
- Go to the cart and remove product, check alert classes.

@hsankala FYI
@justinkruit IYLIMI


justinkruit:
Will test this PR in the weekend. I want to make sure it doesn't break when the theme has Gutenberg usage

crftwrk:
I can confirm that the alerts have been fixed in WC 8.6. Check https://bootscore.me/shop/ alerts in offcanvas cart, cart and checkout.

justinkruit:
> I can confirm that the alerts have been fixed in WC 8.6. Check https://bootscore.me/shop/ alerts in offcanvas cart, cart and checkout.

Also made sure it's working when users have enabled Gutenberg in their theme.

justinkruit (approved):

PR #682, opened 2024-02-16
Move assets to assets folder

### LINKED ISSUE #669: Create assets folder
The amount of folders in the main folder is growing and growing. Do we want to switch to using an /assets folder?
That way we can put /css, /img, /js, /scss and /fontawesome in folder.

This can be a breaking change in our new big version.

justinkruit:
This PR will move CSS, SCSS, JS, img and FontAwesome to a separate assets folder. This will help keep the folder structure cleaner.

As a small extra I've also renamed the .map file of the css to be main.css.map instead of main.map. This is in line with how it should be saved and is preferred by IDEs.

This PR closes #669


crftwrk (approved):
Looks good. Before merging, I did the same for the child https://github.com/bootscore/bootscore-child/pull/41. May have a quick look?

crftwrk (changes requested):
Found a bug that source map is not linked correct in compiled CSS:

```
/*# sourceMappingURL=/wp-content/themes/bootscore-child-dev-main/assets/css/main.map */
```

https://github.com/bootscore/bootscore/blob/e4e2e2720f6536108111ddfecfe57bc521e58de3/inc/scss-compiler.php#L54


`'map'` should be replaced by `'css.map'`?

justinkruit:
> Found a bug that source map is not linked correct in compiled CSS:

Good catch! Fixed it in abe3d9d

crftwrk (approved):
Great, may have a quick view to the child's assets folder https://github.com/bootscore/bootscore-child/pull/41?

PR #684, opened 2024-02-17
Make breadcrumb always responsive

crftwrk:
Until now, our breadcrumb was only horizontally scrollable on mobile but not on lg devices. I have many clients who uses long categories and titles which breaks breadcrumb layout on larger devices. This PR make the breadcrumb scrollable on all devices simply with Bootstrap utilities.

Check https://dev.bootscore.me/2009/07/02/edge-case-many-categories/

| Before  | After |
| before  | after  |

@justinkruit IYLIMI


justinkruit:
While it does look better since it's now one line, I also find it to be less clear that it's scrollable. Also unless you have a horizontal scroll wheel on your mouse, you can't actually scroll. Anything we can do about that?

crftwrk:
We can remove `scrollbar-none` class, then there are scrollbars added by the browser.

crftwrk:
Roughly removed the scrollbar-none class on the dev site. Now there are scrollbars and they look good on my Apple devices and appears on hover. But I do not know how they look on Windows or Linux. Appear they on hover or are they always visible? What is if breadcrumb has short content? @justinkruit 

- https://dev.bootscore.me/2009/07/02/edge-case-many-categories/
- https://dev.bootscore.me/2009/08/06/edge-case-no-content/

justinkruit:
> But I do not know how they look on Windows or Linux.

Unfortunately, it doesn't look as good on Windows.
image

crftwrk:
Quickly changed it. please check both pages:

- https://dev.bootscore.me/2009/07/02/edge-case-many-categories/
- https://dev.bootscore.me/2018/11/03/block-image/

justinkruit:
Could we add the following to the nav?
```css
scrollbar-color: rgba(var(--bs-body-color-rgb),.16) transparent;
```
It will make the navbar look way cleaner in this instance.

| Before | After |
|--------|--------|
| image| image|

justinkruit:
In that case no scrollbar is rendered because it's `auto`.

image

crftwrk:
Ok, that looks fine. I will update this PR soon.

justinkruit (approved):
All looks good now 👍

PR #685, opened 2024-02-20
Rename  bscore to  bootscore

crftwrk:
This PR replaces all `_bscore*_.scss` with `_bootscore_*.scss` and `--bscore` variables with `#{$prefix}`. I think this is better to achieve a consistent line in the theme. Similar to function names.

However, I'm not very happy with those color `color: var(--#{$prefix}wc-form-invalid, var(--#{$prefix}danger));` management, but this can be improved later.

Child PR https://github.com/bootscore/bootscore-child/pull/43

@justinkruit IYLIMI


justinkruit:
> and `--bscore` variables with `#{$prefix}`

This will mean that we use the Bootstrap prefix, and won't have a difference between bootscore variables and bootstrap variables. Is that alright?

crftwrk:
> This will mean that we use the Bootstrap prefix, and won't have a difference between bootscore variables and bootstrap variables. Is that alright?

Correct

justinkruit (approved):
Tested it, and all looks good 👍

PR #686, opened 2024-02-21
Update libraries

crftwrk:


justinkruit (approved):
Also seems to work all fine for me 👍

PR #703, opened 2024-02-26
WC Ajax Mini Cart Quantity Buttons

anisur2805:
Added the Ajax functionality for adding products to cart.

Closes https://github.com/bootscore/bootscore/issues/631


crftwrk:
Hi Anisur,

thank you, we are on a good way! 

### mini-cart.php

#### 1. Check if product is only 1 in stock left or sold individually

It should be checked if products are out of stock or sold individually. Please add this 3 products to the cart:

- https://dev.bootscore.me/product/album/
- https://dev.bootscore.me/product/beanie/ (only 1 in stock left)
- https://dev.bootscore.me/product/hoodie-with-pocket/ (sold individually)

After adding these products, go to cart page. Quantity for "Beanie" and "Hoodie with pocket" should be disabled.

WooCommerce uses following code to check this in cart.php:

```php
if ( $_product->is_sold_individually() ) {
    $min_quantity = 1;
    $max_quantity = 1;
} else {
    $min_quantity = 0;
    $max_quantity = $_product->get_max_purchase_quantity();
}
```

#### 2.  If possible adding the quantity by a hook instead adding them direct to the template

Something like this in **ajax-add-to-cart.php**:

```php
add_filter('woocommerce_widget_cart_item_quantity', 'add_minicart_quantity_fields', 10, 3);
function add_minicart_quantity_fields($html, $cart_item, $cart_item_key) {
    $product_price = apply_filters('woocommerce_cart_item_price', WC()->cart->get_product_price($cart_item['data']), $cart_item, $cart_item_key);

    return woocommerce_quantity_input(array('input_value' => $cart_item['quantity']), $cart_item['data'], false) . $product_price;
}
```

This is important if user disables the entire **ajax-add-to-cart.php** in child by function:

```php
// Disable AJAX Cart
function register_ajax_cart() {
}
add_action('after_setup_theme', 'register_ajax_cart');
```

The reason for that is that quantity should not appear in mini-cart if user disabled the the file.

### woocommerce.js

All additioinal code should be placed within a function in **ajax-add-to-cart.php** as well. Code should not load if users disables **ajax-add-to-cart.php**.

In conclusion everything what is is related to ajax functionality should be placed in **ajax-add-to-cart.php**, to simply disable it in one single file.

**wc-minicart.php** and **wc-qty-btn.php** looks good to me!

Thanks in advance

crftwrk:
Hi Anisur,

yes! Just two tiny things:

- If I disable ajax cart, then quantity buttons in product page do not work, because JS is now in ajax-add-to-cart.php. In this case I prefer to move the JS part back to woocommerce.js as you did it before.
- After that, please remove then all commented code.

Last thing: Is it possible for you to update your fork, open a new PR and commit to https://github.com/bootscore/bootscore/tree/v6.0.0-beta1 branch? That's because many things happen since you started here and this will allow me to merge without asking for a review. This is not a must have, but will make things happen much more faster.

Thank you!

justinkruit:
> Last thing: Is it possible for you to update your fork, open a new PR and commit to https://github.com/bootscore/bootscore/tree/v6.0.0-beta1 branch?

This PR can be changed to merge into the v6 branch instead of the main branch. On GitHub when you edit the whole PR, and probably via git commands too.

crftwrk:
@justinkruit Got it

crftwrk (approved):
👍

PR #707, opened 2024-02-27
Add more filters to header

crftwrk:
```php
/*
 * Header position and bg
 */
function header_bg_class() {
  return "sticky-top bg-body shadow";
}
add_filter('bootscore/header_class', 'header_bg_class', 10, 2);
```

```php
/*
 * Header buttons
 */
function header_button_class() {
  return "btn btn-link ps-2 pe-0 text-decoration-none text-body-secondary";
}
add_filter('bootscore/header_button_class', 'header_button_class', 10, 2);
```

### DIFF
--- a/header.php
+++ b/header.php
@@ -36,7 +36,7 @@ defined('ABSPATH') || exit;
     <?php dynamic_sidebar('top-bar'); ?>
   <?php endif; ?>  
 
-  <header id="masthead" class="site-header sticky-top bg-body-tertiary">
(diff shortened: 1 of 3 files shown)


justinkruit:
Can you change the button class filter?

- rename it to bootscore/header/toggler_class
- add which button it is as extra argument

crftwrk:
How do you mean this exactly? May you want quickly do a PR?

justinkruit:
Sure, I can do that when I get home

PR #709, opened 2024-02-27
Rename some new filters

justinkruit:
- Renamed `bootscore/header_button_class` to `bootscore/header/button_class` and added which button it is.

I'm also now debating, what would be better filter names?
| Current | Idea |
|--------|--------|
| `bootscore/header_button_class` | `bootscore/class/header_button` | 
| `bootscore/header_class` | `bootscore/class/header` | 
| `bootscore/container_class` | `bootscore/class/container` | 

This idea is coming from the principle of merging things that are alike, like the filter `bootscore/offcanvas/navbar/title`. With the other filters, we do nothing more than managing their classes. So a `bootscore/container/class` would imply more things happening with the container. I'm open to feedback @crftwrk

### DIFF
--- a/404.php
+++ b/404.php
@@ -14,7 +14,7 @@ defined('ABSPATH') || exit;
 
 get_header();
 ?>
-  <div id="content" class="site-content <?= apply_filters('bootscore/container_class', 'container', '404'); ?> <?= apply_filters('bootscore/content/spacer_class', 'pt-4 pb-5', '404'); ?>">
+  <div id="content" class="site-content <?= apply_filters('bootscore/class/container', 'container', '404'); ?> <?= apply_filters('bootscore/content/spacer_class', 'pt-4 pb-5', '404'); ?>">
     <div id="primary" class="content-area">
 
       <main id="main" class="site-main">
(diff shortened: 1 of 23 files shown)


crftwrk:
Your idea sounds good, I‘m in. We need at the end for each filter an example for the docs.

justinkruit:
Here is a list with the changed filter names:
| Old                               | New                               |
|-----------------------------------|-----------------------------------|
| bootscore/container_class         | bootscore/class/container         |
| bootscore/main/col_class          | bootscore/class/col/main          |
| bootscore/footer_class            | bootscore/class/footer            |
| bootscore/footer_info_class       | bootscore/class/footer_info       |
| bootscore/to_top_button_class     | bootscore/class/to_top_button     |
| bootscore/header_class            | bootscore/class/header            |
| bootscore/header/button_class     | bootscore/class/header_button     |
| bootscore/sidebar/col_class       | bootscore/class/col/sidebar       |
| bootscore/sidebar/toggler_class   | bootscore/class/sidebar/button    |
| bootscore/sidebar/offcanvas_class | bootscore/class/sidebar/offcanvas |

PR #716, opened 2024-03-02
[Fix] Mini Cart Currency Symbol Position

anisur2805:
Hey @crftwrk,
I have checked this currency position, and it is not updated according to woo-commerce backend settings without product quantity update or added to the cart. I have RnD about it and found a notice in the admin panel. I have checked more than 3/4 industry's top themes which have done the same way to update their quantity. As far as my RnD it's a default behaviour of WooCommerce. 

Please share your thoughts.

image

### DIFF
--- a/woocommerce/cart/mini-cart.php
+++ b/woocommerce/cart/mini-cart.php
@@ -70,7 +70,13 @@ do_action('woocommerce_before_mini_cart'); ?>
                   </a></strong>
               <?php endif; ?>
               <div class="item-quantity">
-                <?php echo wc_get_formatted_cart_item_data($cart_item); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped 
+                <?php echo wc_get_formatted_cart_item_data($cart_item); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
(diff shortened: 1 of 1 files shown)


justinkruit:
This is not the correct way to render prices in WooCommerce. The function wc_price does this formatting for you, and we should actually use that.

justinkruit (changes requested):
Use wc_price to format the price (see above comment).

crftwrk:
@justinkruit What's wrong here? Works fine? Is this something what we can change later by ourself easily?

justinkruit:
> Is this something what we can change later by ourself easily?

We can, but imo we should fix it before the v6 release.

> What's wrong here? Works fine?

It works, but it's not the correct way to render a price in Woocommerce. Here it's manually built, while woocommerce has a function for this built in.

crftwrk:
@justinkruit can you provide a quick code example how it should look like?

anisur2805:
@justinkruit here is the explanatory video I have recorded a [video](https://www.awesomescreenshot.com/video/25458233?key=c422e0f320100acb89a98649f750113c) for you. Please have a look. And I will fix the price display code as you have suggested me within very soon. 

Thank you.

PR #718, opened 2024-03-04
Fix dynamic currency position

anisur2805:


crftwrk (approved):
@justinkruit are you fine here?

justinkruit:
I want to make sure this is correctly implemented because of the possibility of customizations on custom webshops. 
https://github.com/bootscore/bootscore/blob/0b1d861ee03329bcc6c5c89620367a0cd21c6613/woocommerce/cart/mini-cart.php#L43

This line seems to be unused now, so the filter `woocommerce_cart_item_price` does nothing here.

I'll look closer into it later if that's alright, when I'm not at work hahaha

woocommerce/cart/mini-cart.php
```diff
@@ -72,7 +72,7 @@
               <div class="item-quantity">
                 <?php echo wc_get_formatted_cart_item_data($cart_item); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped 
                 ?>
-                <?php echo apply_filters('woocommerce_widget_cart_item_quantity', '<span class="quantity">' . sprintf('<span class="qty_text">%s</span> &times; %s', $cart_item['quantity'], $product_price) . '</span>', $cart_item, $cart_item_key); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped 
+                <?php echo apply_filters('woocommerce_widget_cart_item_quantity', '<span class="quantity">' . sprintf('<span class="qty_text">%s</span> &times; %s', $cart_item['quantity'], wc_price( $cart_item['line_total'] ) ) . '</span>', $cart_item, $cart_item_key); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
justinkruit, line 75: @anisur2805 Can you please give an explanation as to why this line has to be changed? I did some testing, and it has the same result for me as before the change. The only thing I've noticed making a change is the one you made on line 97.

anisur2805: @justinkruit as earlier you mention to use `wc_price()` that's why I updated the line and it's working fine. Please let me know if you have any suggestions. Thanks.

crftwrk: @justinkruit is it an option for you to add the previous line as a comment and review it before v6 release? It works fine for now and I think we should merge to keep workflow going on.

justinkruit: @anisur2805 seems like this line isn't needed. The current line using $product_price seems to be already formatted. This line can be reverted. The other line can stay with the change.

@crftwrk this will be an easy fix and merged with V6 👍

crftwrk: @justinkruit are you fine to merge now here?

PR #729, opened 2024-03-08
Improve the mini-cart content

crftwrk:


username0136:
wow this is real great :fire:

Issue #621, opened 2023-11-11
Refactor the logo


crftwrk:
Many brands have transitioned from using a combination of image, word, and slogan to adopting a simple icon that serves as a representative symbol for the entire brand (ignoring the Twitter X).

| Beginning  | Today |
| ------------- | ------------- |
| <img src="https://github.com/bootscore/bootscore/assets/51531217/81fbd9a5-83bf-452a-87d6-606b1d88c212" width="100">  | <img src="https://github.com/bootscore/bootscore/assets/51531217/9a3f0b1d-031b-4e75-ac2f-5a24ffb06bc8" width="100">  |
| <img src="https://github.com/bootscore/bootscore/assets/51531217/8886256e-3980-404a-8c49-3a9850e0fb91" width="250">  | <img src="https://github.com/bootscore/bootscore/assets/51531217/aee5e08f-a508-45eb-857c-2772bde7beea" width="100">  |
| <img src="https://github.com/bootscore/bootscore/assets/51531217/ba163532-59ca-4085-a76d-2ca8afc43210" width="250">  | <img src="https://github.com/bootscore/bootscore/assets/51531217/ba6a56fa-6f6e-467a-afcf-547d1464b885" width="100">  |

With this in mind, I have reevaluated our brand.

## Status quo

We currently have a charming wordmark inspired by the 'Skype' style that has served us well over the years.

| Mobile  | Desktop |
| ------------- | ------------- |
| <img src="https://github.com/bootscore/bootscore/assets/51531217/9b4df1ab-7ab7-48c8-acf5-9d77c84c5412" height="50">  | <img src="https://github.com/bootscore/bootscore/assets/51531217/f29284fe-8534-4341-9717-f38ffdb45d9e" height="50">  |

However, as times evolve, I believe it's time to elevate our branding to the next level.

## Icon

**BootScore** is a combination of **Boot**strap and the Under**score**s theme boilerplate. Breaking these two words down to the root leads me to create something very simple, like `b_`, which represents the boot and the underscore. Moving the `_` below a capital `B` and using it as the sole for the boot simplifies it further and communicates two things in one icon using the golden ratio.

| B + _  |
| ------------- |
| <img src="https://github.com/bootscore/bootscore/assets/51531217/a3af151a-5e2c-4006-9027-dc54e5132393" height="200">  |

Placing the icon in a square allows an outline and a bold variant.

[...]

PR #732, opened 2024-03-11
Refactor the logo and add a utility to hide elements on data-bs-theme=""

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/621


justinkruit:
Fun fact: in an svg, you could change the color depending on the dark mode preference.

https://stackoverflow.com/questions/67187091/creating-svg-that-appears-black-in-light-mode-and-light-in-dark-mode (best example that I could find with a quick Google search)

crftwrk:
Yes, I know ;-) But this will be hard for users, so two logos seems to be better here.

PR #750, opened 2024-04-17
Experimental: Skip cart page

crftwrk:
I do not want to merge this, just collecting ideas


stan1781:
Why not just add a “buy now” button that directly takes you to the “checkout” page upon clicking!

crftwrk:
> Why not just add a “buy now” button that directly takes you to the “checkout” page upon clicking!

This will not work for all product types, goes too deep into WooCommerce and will create issues for the most users. Check the new PR https://github.com/bootscore/bootscore/pull/755

PR #755, opened 2024-04-26
Skip cart page

crftwrk:
This is an idea for later. Because the new AJAX cart is a full working one now, we can hide the "View cart" button, skip the entire cart page and redirect to checkout

Use a filter to restore default cart and button:

```php
/**
 * Disable skip cart and enable the cart page
 */
add_filter('bootscore/skip_cart', '__return_false');
```

### DIFF
--- a/woocommerce/inc/wc-skip-cart.php
+++ b/woocommerce/inc/wc-skip-cart.php
@@ -1,0 +1,47 @@
+<?php
+
+/**
+ * WooCommerce Skip Cart Page
+ *
+ * @package Bootscore 
+ * @version 6.0.0
+ */
+
+
+// Exit if accessed directly
+defined('ABSPATH') || exit;
+
+
+
+/**
+ * Mini cart widget buttons
+ */
+remove_action('woocommerce_widget_shopping_cart_buttons', 'woocommerce_widget_shopping_cart_proceed_to_checkout', 20);
+add_action('woocommerce_widget_shopping_cart_buttons', 'bootscore_widget_shopping_cart_proceed_to_checkout', 20);
+
+function bootscore_widget_shopping_cart_proceed_to_checkout() {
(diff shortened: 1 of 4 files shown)


Remzi1993:
Nice! 👍😍

woocommerce/inc/wc-skip-cart.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Skip cart redirecting to checkout
+ */
+function bootscore_skip_cart_page_redirection_to_checkout() {
+
+  // If is cart page, redirect checkout.
+  if( is_cart() )
+    wp_redirect( wc_get_checkout_url() );
+}
+add_action('template_redirect', 'bootscore_skip_cart_page_redirection_to_checkout');
```
justinkruit, line 47: I think we should add to our documentation that if you want the redirect to be removed, you should do a `remove_action`. Agreed @crftwrk?

woocommerce/inc/wc-skip-cart.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Hide WooCommerce Mini-Cart "View Cart" Button
+ */
+remove_action('woocommerce_widget_shopping_cart_buttons', 'woocommerce_widget_shopping_cart_button_view_cart', 10);
+add_action('woocommerce_widget_shopping_cart_buttons', 'bootscore_remove_view_cart_minicart', 10);
+ 
+function bootscore_remove_view_cart_minicart() {
+  remove_action( 'woocommerce_widget_shopping_cart_buttons', 'woocommerce_widget_shopping_cart_button_view_cart', 20 );
+}
```
justinkruit, line 35: This seems to have the same `remove_action` 2 times?

justinkruit (changes requested):
All seems fine, just have 2 remarks.

justinkruit (approved):

Issue #442, opened 2023-04-03
[Feature Request] editor-style.scss


Effectiva:
### Describe the feature that you'd like

Hi,

In the folder scss i've added the file editor-style.scss, but its not compiling to editor-style.css?
My other main files main.scss does work?

### Checklist

- [X] I have searched the issues and this feature has not been requested.

crftwrk:
Hi, 

unfortunately you cannot use SCSS compiler for editor https://github.com/orgs/bootscore/discussions/424. You have to enqueue an usual CSS file https://codex.wordpress.org/Editor_Style.

Effectiva:
Hi, 

Yes thats right, but when I try to enqueue editor-style.css it works, however the compiler doesn't compile editor-style.scss to editor-style.css

crftwrk:
> however the compiler doesn't compile editor-style.scss to editor-style.css

Yes, that's not possible https://github.com/orgs/bootscore/discussions/424#discussioncomment-5228488

Effectiva:
Ah ok, that would be nice if its possible in the future. Clients do like to have a styled backend (blocks). I did try to work with wp-scss but that doesn't work/gives errors (I did disable the built in compiler)

Issue #55, opened 2021-12-13
Bootstrap color palette to Gutenberg block editor 


iwasthesword:
I propose we add custom Bootstrap palette to be added to the block editor. [Ref](https://github.com/understrap/understrap/issues/1295) 
If you are using the block editor this is a feature that would help A LOT to keep consistency across your site.

Since we have pre-processing I'm guessing it could be done as in the reference, but since I'm not familiar with the process in here I'd like to hear your opnions before diving in. What are your opinions on this? How should I go about implementing it? Feel free to join in

crftwrk:
I think it's a good idea to add some basic support for the block editor. We're happy if you want to contribute. The process here is simple. Just fork the theme repo, do your changes and create a Pull request.

iwasthesword:
I think that it wouldn't be that hard, plus it would do ground work for more theme customization. Since `theme.json` is available it would make it a lot easier to implement and not depend on the PHP `add_theme_support` way of doing it.

The only thing we would need to make is the for `scss-compiler.php` file to generate the `theme.json` in the root dir with all the colors as specified in here.

I'm not sure SCSS Compiler CAN compile to .json buit looking at the SCSS code and generated CSS file I'd say the first part of the `_root.scss` has everything we need to do this, list all the colors:

``` 
@each $color, $value in $colors {
    --#{$variable-prefix}#{$color}: #{$value};
}
```

Any idea how I would about doing this? I'm not familiar with SCSS compiler and I'm having a hard time figuring it out how to add just the colors to the .json

justinkruit:
Building support in the scss is quite easy, using the example you posted. Unfortunately, the json/PHP is a bit harder. The compiler can't just read the variables and write json for it. This would basically require all the scss variables to be declared in PHP in some way, so both the compiler and a file writer for json can both use them.

In understrap they use gulp to generate the code, unfortunately in the current system of bootScore this won't fit, as we mainly focus on not needing additional tools. I'll take a closer look at the code on understrap soon, to see if I can get some inspiration.

The only way I can imagine this working the best if we give people instructions on how to edit the theme.json.

iwasthesword:
I've did a bit of research and got a working prototype:

- It uses an external tool https://github.com/sabberworm/PHP-CSS-Parser
`composer require sabberworm/php-css-parser` to add the library
- After the SCSS Compiler does all the work, it will load the bootstrap.min.css and parse the variables.
[...]

PR #757, opened 2024-04-28
Register editor style.css, add theme.json and create block patterns

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/55
Closes https://github.com/bootscore/bootscore/issues/442

https://youtu.be/GrLLuyhJP1g (outdated)

This is a good one addressed to v6 stable. Compiled `main.css` is now added to the editor and Bootstrap components can now be added without using a plugin.

This PR seems big, but it's tiny because magic happens only in two files. Means that not every file must be reviewed, overall functionality is enough.

- Registers compiled `main.css` as `editor-style.css`. Bootstrap styles are now applied to the GB editor
- Added `theme.json` with color variables
- Created some patterns like cols and buttons. Click the + button on top left to insert a new block > Patterns > Bootscore > choose block pattern

| Backend  | Frontend |
| backend  | frontend  |

### editor-style

Adds the compiled `main.css` to the GB editor:

https://github.com/bootscore/bootscore/blob/8d637abcbf15ea8d921b5fd5de699b15a6a09e54/inc/enqueue.php#L59-L60

### theme.json

The `theme.json` is added to the root folder and needed to remove the inner `div` in a GB group-block. Because editor-style is now enqueued, there are no issues with broken fonts etc. as mentioned in https://github.com/bootscore/bootscore/issues/55#issuecomment-1103956863. Defined fonts in Bootstrap are assigned to editor as well.

`theme.json` usually uses fixed HEX or RGB values, but here variables `var(--bs-primary)` etc. are used to get defined colors from `main.css`. This works fine for the editor but unfortunately not in the sidebar where colors can be selected, because `main.css` is only enqueued to the main editor fields.

For this purpose I enqueued `main.css` a second time to the general admin area if it is the post edit screen. This leads to minor style issues like the size of the generated preview `iframe` in patterns. That's why there is a new `admin.css` enqueued to the backend editor which fixes that. Think this `admin.css` is a good idea, because we can add for example a border to group-blocks `div` to make them visible and better to select in the backend later.

https://github.com/bootscore/bootscore/blob/8d637abcbf15ea8d921b5fd5de699b15a6a09e54/inc/enqueue.php#L65-L67

### Patterns

Finally with stuff above, I've added some block patterns, which can be found in folder `patterns`.

https://github.com/bootscore/bootscore/blob/8d637abcbf15ea8d921b5fd5de699b15a6a09e54/inc/theme-setup.php#L82-L85

Once used, patterns are stored into the database, means that we can add/change/remove patterns in future without having any impact to existing sites. Created already some landing pages on live-sites and they work well. Even bootscore.me's frontpage is already rebuild with this patterns:

- 2 Alerts
- 3 Buttons
- 3 cards and a card grid with pricing example
- container
- 4 types of cols
- 3 Heroes
- 4 prebuilt full-width CTA sections to use on the `page-blank.php`

@justinkruit IYLIMI

p.s. you can preview them here https://bootscore.me/documentation/block-patterns/ (private)


justinkruit (commented):
Are you aware that loading Bootstrap css like that, also changes things like the headings etc? Here the default Bootstrap things will also be applied.

We can make a separate css file which doesn't have the reboot styling so we only focus on the components.

inc/enqueue.php
```diff
@@ -51,6 +51,26 @@ function bootscore_scripts() {
...
+function bootscore_add_editor_styles() {
+  add_theme_support('editor-styles');
+  add_editor_style('assets/css/main.css');
+
+  // Check if the current page is the Gutenberg editor and enqueue CSS to tha main admin area to use variables in theme.json
+  $screen = get_current_screen();
+  if ($screen && $screen->is_block_editor) {
+    wp_enqueue_style('main-editor-style', get_stylesheet_directory_uri() . '/assets/css/main.css', array(), '1.0', 'all');
+    // Fix minor CSS issues in the backend editor screen
+    wp_enqueue_style('admin-style', get_template_directory_uri() . '/assets/admin/css/admin.css', array(), '1.0', 'all');
```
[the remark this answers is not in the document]
crftwrk: > We can make a separate css file which doesn't have the reboot styling so we only focus on the components.

Yes, any idea?

> Shouldn't we use filemtime here for the version?

Yes, we can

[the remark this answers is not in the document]
justinkruit: > Yes, any idea?

If we make a new scss file, we exclude some of the parts of Bootstrap easily. Just need to be aware that this can raise the loading times because more scss is compiled, so we need to add a way that it'll skip some compiler checks... May need to expand the SCSS compiler before we can achieve that.

crftwrk:
@justinkruit following changes:

- Thanks to the new compiler, I added a new `editor.scss` which only compiles the `_root.scss` for the color palette in GB sidebar. Far better than previous solution.
- This file gets compiled every time when `main.css` is compiled. No need to add a `filemtime` here, because backend styles are not cached.
- To enjoy this feature the `editor.scss` must be added to child as well https://github.com/bootscore/bootscore-child/pull/51, otherwise color picker items in GB sidebar are grey.

Tested it with main- and child-theme and at some live sites as well. Think this is good.

crftwrk:
WP 6.6 made changes in pattern margins if group is set to `is-layout-flow`. Fixed in https://github.com/bootscore/bootscore/pull/811

inc/scss-compiler.php
```diff
@@ -212,10 +212,19 @@ function bootscore_child_has_scss() {
...
+                      ->addModifiedCheck(get_template_directory() . '/assets/scss/bootstrap/bootstrap.scss', false)
+                      ->compile();
+  
+  // Compile the editor.scss file
+  $scss_compiler_editor = new BootscoreScssCompiler();
+  $scss_compiler_editor ->scssFile('/assets/scss/editor.scss')
+                        ->cssFile('/assets/css/editor.css')
+                        ->addModifiedCheckTheme()
+                        ->addModifiedCheck(get_stylesheet_directory() . '/assets/scss/main.css', false)
+                        ->compile();
```
justinkruit, line 229: I wouldn't do it like this. This means every time main is compiled (which is each time in dev mode), it also compiles the editor.css. I would recommend only checking the bootstrap files and the variables. It may also be smart to add an additional function to the compiler to skip the env check.

justinkruit (changes requested):

inc/scss-compiler.php
crftwrk: Agree, want to change this quickly? I‘m not in front of my machine.

justinkruit: Already got it done 😉

crftwrk: Perfect, merge?

justinkruit (approved):
Looks all good to me 👍

justinkruit:
> Can we go for a release on Thursday?

Is Friday alright? Thursday won't play nice with my work schedule.

crftwrk:
Tomorrow fine as well? Friday I go to holiday.

justinkruit:
In that case Thursday afternoon/evening is the best option, tomorrow I will barely be at home.

crftwrk:
> Thursday afternoon/evening

is absolutely fine for me. But we can do it next week as well if you prefer that.

PR #776, opened 2024-05-20
A better .gitignore and dynamic footer columns based activated footer widgets

Remzi1993:
Generated a better .gitignore using gitignore.io and made the footer columns dynamic based on activated footer widgets.

Basically checking how many footer widgets are active:
```
function bootscore_count_active_footer_widgets(): int {
    $count = 0;
    for ($i = 1; $i <= 4; $i++) {
        if (is_active_sidebar('footer-' . $i)) {
            $count++;
        }
    }
    return $count;
}
```
And then base the columns on that:
```
function bootscore_get_footer_widget_class(): string {
    $widget_count = bootscore_count_active_footer_widgets();
    return match ($widget_count) {
        1 => 'col-12',
        2 => 'col-6',
        3 => 'col-4',
        default => 'col-3',
    };
}
```


crftwrk:
Hi Remzi,

Thank you for this PR, we're are always happy if someone wants to contribute! But I have some questions.

Why do we need this massive `.gitignore`? Never seen this before even in large projects.

About the footer. We had already some thoughts on this but we always end up as it is. Because since v6 we have those sleek [filters](https://bootscore.me/documentation/filters/#Footer_1-4_columns) which allows to edit columns in a simple way. Additionally you can use a widget in footer 1, 3, 4 and leave footer 2 empty as a spacer. This would not be possible with dynamic widgets.

Just for upcoming PR's:

- Please use default 2 empty spaces for formatting instead 4 or tab.
- Functions should be added in a functions file in folder `inc` and not in a template file.
- Please do not pack different types of changes into one PR. Split them into two, one for the footer and one for the `.gitignore`. This helps us to keep track of things and, if necessary, to be able to reverse a PR.

I set the PR to draft so it wouldn't be merged accidentally. Let's wait for @justinkruit for his thoughts.

Remzi1993:
I'm going to answer your questions step by step:

> Why do we need this massive `.gitignore`? Never seen this before even in large projects.

If you look at: https://www.toptal.com/developers/gitignore?templates=linux,macos,windows,composer,wordpress,intellij+all,phpstorm+all,visualstudiocode,node,yarn
You will see I have ignored all operating systems: Windows, Linux and Mac and the 3 populair IDE's (IntelliJ IDEA, PhpStorm and Visual Studio Code), I have also ignored WordPress files, Composer, Node and Yarn.

These are either all the things you use or other devs could use like IntelliJ or VS Code. My advice would be to either generate your .gitignore and include als much as you can then too little. Because you don't want to include things which take up space which can recreated all the time. BTW, I use PhpStorm for this project, but use IntelliJ and VS Code for different things.

> Additionally you can use a widget in footer 1, 3, 4 and leave footer 2 empty as a spacer. This would not be possible with dynamic widgets.

I think this is possible with my code, you would simply have 2 columns instead of 1. Have you test this out? This code has nothing to do with widgets but only show a number of columns depending how may footer widgets are active (used).

> Just for upcoming PR's:
> 
> * Please use default 2 empty spaces for formatting instead 4 or tab.
> * Functions should be added in a functions file in folder `inc` and not in a template file.
> * Please do not pack different types of changes into one PR, split them into two. One for the footer and one for the `.gitignore`. This helps us to keep track of things and, if necessary, to be able to reverse a PR.

Will keep this in mind, could you please put this in README or a separate CONTRIBUTING file? This would be easier if someone wants to contribute.

crftwrk:
Got it with .gitignore and agree. But when we open this box, this must be added to all child and plugin repos as well. This lines can be deleted, because this files have been moved to assets folder:

```
css/main.css
css/main.map
```

For the footer widgets, yes its possible to create empty cols but it adds the `mb-3` class to it because an empty `<div>` is recognised as a widget. It is not responsive in your example too. However, as I mentioned something like this we had in mind as well but always end up as it is now. I do not really get warm with it and we should rethink this in v6.1. Focus is now on v6 stable.

> Will keep this in mind, could you please put this in README or a separate CONTRIBUTING file? This would be easier if someone wants to contribute.

Absolutely! A `CONTRIBUTING.md` is something we should add here. At the moment we are struggling with stuff for v6 stable which have a higher priority. If you are able to create one, we will greatly appreciate it.

Remzi1993:
> yes its possible to create empty cols

My apologies, that's indeed not a nice solution. I did a quick and dirty fix for a client. The columns should only be created if there are active widgets in Footer 1, Footer 2, etc....

I can't look at it at the moment because I need to finish 2 other school projects first. Maybe I can come up with a solution when the academic year ends and when the summer holiday starts.

Issue #774, opened 2024-05-19
my-account/edit-address/


dimage:
https://site.com/my-account/edit-address/

Woocommerce>>>Settings>>Shipping>>>Shipping settings>>>Shipping destination>>>Force shipping to the customer billing address

![Screenshot from 2024-05-19 20-01-14](https://github.com/bootscore/bootscore-child/assets/10204962/671df9a1-8ed8-4455-80a7-9cea2c72301f)

PR #788, opened 2024-05-31
Minor fixes to v6 stable

crftwrk:
Files in folder `languages` can be ignored. The rest are just minor fixes.

- Update `woocommerce/my-account/navigation.php` (WC 9)
- Increase the `z-index` of the `to-top` button so that it is not overlapped by the prev/next swiper buttons
- Fixes https://github.com/bootscore/bootscore/issues/774
- Improves the product gallery styles
- Added a class to full-width-image templates title
- Update the translation template
- Fix line-height in latest comments block
- Fixes https://github.com/bootscore/bootscore/issues/795

@justinkruit IYLIMI. After this I have just one more PR to add an auto-updater, and then that's it.


justinkruit (approved):
Looks fine to me 👍

Issue #182, opened 2022-07-01
Auto update?


vahidalvandi:
Do this theme support auto update ?

crftwrk:
Not yet, because we still doing some breaking changes. When there are absolutely no breaking changes anymore, we can add an auto updater. Maybe 5.3.0.0.

vahidalvandi:
Thank you

crftwrk:
Discussed here https://github.com/bootscore/bootscore/issues/30

PR #797, opened 2024-06-21
Include auto updater

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/182

This is the last PR for v6 stable and includes the auto-updater https://github.com/YahnisElsts/plugin-update-checker. I think we are ready now to do this step, because I have no idea for breaking changes anymore. It feels good and forces users to update their theme by a single click.

Created some releases in a private repo for testing:

- Download and install [bootscore.zip](https://github.com/user-attachments/files/15936003/bootscore.zip)
- Go in backend to Dashboard > Updates and update theme to v6.2.

@justinkruit IYLIMI


justinkruit (approved):
Looks good to me!

PR #802, opened 2024-06-28
Rewrite SCSS compiler to a class

justinkruit:
This PR was made in response to a comment on #757.

Before, the entire SCSS compiler only supported one file, the main.scss file. Now this new class supports a more dynamic method of compiling SCSS. The original `bootscore_compile_scss` function is still being used, it has just been renewed to fully use the new class.

I've tested it on both the normal and child theme, and everything seems to work just like it had done before, with no compromises.

@crftwrk would love your honest feedback!

### DIFF
--- a/inc/scss-compiler.php
+++ b/inc/scss-compiler.php
@@ -1,9 +1,9 @@
 <?php
 
 /**
- * Functions to compile scss in the theme itself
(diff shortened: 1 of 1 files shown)


crftwrk (approved):
Looks good!

Issue #805, opened 2024-07-01
Source map not linked correctly in dev mode


crftwrk:
### Describe the bug

Since https://github.com/bootscore/bootscore/pull/802 developer mode not works correctly. 

### Steps to reproduce

Add 

```php
define('WP_ENVIRONMENT_TYPE', 'development');
```

to `wp-config.php`, reload the frontend and inspect the elements. Source map is not linked correctly in the `main.css`:

```css
/*# sourceMappingURL=/mywp//assets/css/main.css.map */
```

instead

```css
/*# sourceMappingURL=/mywp/wp-content/themes/bootscore-child/assets/css/main.css.map */
```

Is it something here?

https://github.com/bootscore/bootscore/blob/53db475df0f29680f285a0e51f4143b35be154d8/inc/scss-compiler.php#L115-L127

@justinkruit may you have a look to it?

PR #806, opened 2024-07-02
Fix path to SCSS source map

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/805

https://github.com/bootscore/bootscore/pull/802 breaks the path to the `main.css.map` in compiled `main.css` if site is under development. This PR fixes that.

Add `define('WP_ENVIRONMENT_TYPE', 'development');` to `wp-config.php` and inspect frontend styles.

@justinkruit quickly tried to fix that by myself, let me know if this is correct what I'm done.

### DIFF
--- a/inc/scss-compiler.php
+++ b/inc/scss-compiler.php
@@ -114,10 +114,13 @@ class BootscoreScssCompiler {
 
   private function setOutputStyle() {
     if ($this->is_environment_dev) {
+      $source_map_url = site_url('', 'relative') . '/' . ltrim(str_replace(ABSPATH, '', $this->getCssFile()), '/');
+      $source_map_url .= '.map';
+
       $this->compiler->setSourceMap(Compiler::SOURCE_MAP_FILE);
       $this->compiler->setSourceMapOptions([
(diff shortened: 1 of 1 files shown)


inc/scss-compiler.php
```diff
@@ -114,10 +114,13 @@ private function addImportPaths() {
 
   private function setOutputStyle() {
     if ($this->is_environment_dev) {
+      $source_map_url = site_url('', 'relative') . '/' . ltrim(str_replace(ABSPATH, '', $this->getCssFile()), '/');
+      $source_map_url .= '.map';
+
       $this->compiler->setSourceMap(Compiler::SOURCE_MAP_FILE);
       $this->compiler->setSourceMapOptions([
-        'sourceMapURL'      => site_url('', 'relative') . '/' . str_replace(ABSPATH, '', $this->css_file) . '.map',
```
justinkruit, line 119: Is only changing `$this->css_file` to `$this->getCssFile()` not good enough?

justinkruit (approved):
Checked it a bit more, and seems fine. Lets merge

PR #814, opened 2024-07-23
v6 stable

crftwrk:
@justinkruit as usual, there is a drafted release.

PR #816, opened 2024-07-27
Fix redirect checkout if cart is empty

crftwrk:
This fixes a bug that has been overlooked in https://github.com/bootscore/bootscore/pull/755. In my point of view this is an edge case and can be solved in the next maintaining release. This bug happens if you are on checkout page and do nothing more a few days until WC empties the cart by itself by removing the cookie.

Steps to reproduce: 

- Open https://dev.bootscore.me/shop/ twice in two tabs
- Add a product and go to checkout
- Switch to the other tab and remove product from the cart
- Go back to the checkout tab and reload the page

As a result, page stucks because checkout, if cart is empty, redirects to cart page by WC and cart page redirects to checkout by us.

This PR checks if cart is empty and redirects checkout to shop page instead.

### DIFF
--- a/woocommerce/inc/wc-skip-cart.php
+++ b/woocommerce/inc/wc-skip-cart.php
@@ -4,7 +4,7 @@
  * WooCommerce Skip Cart Page
  *
  * @package Bootscore 
- * @version 6.0.0
+ * @version 6.0.1
  */
 
 
--- a/woocommerce/inc/wc-skip-cart.php
+++ b/woocommerce/inc/wc-skip-cart.php
@@ -31,11 +31,20 @@ remove_action('woocommerce_widget_shopping_cart_buttons', 'woocommerce_widget_sh
 
 
 /**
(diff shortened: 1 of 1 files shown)


justinkruit:
Want to do this before #814 ?

crftwrk:
> Want to do this before https://github.com/bootscore/bootscore/pull/814 ?

We can do it now or we can do it in the next maintaining release. So we have something to do in future and can test the auto-updater as well. Both is fine for me. Added a description to describe and reproduce the bug.

justinkruit:
Are we sure we want a redirect to the store in this case? Maybe it's best to just skip the redirect, so you at least see the message of "Empty cart".

crftwrk:
> Maybe it's best to just skip the redirect, so you at least see the message of "Empty cart".

Good idea! See https://github.com/bootscore/bootscore/pull/822

PR #817, opened 2024-07-27
Add Bootstrap copyright banner

crftwrk:
This got lost in https://github.com/bootscore/bootscore/pull/645


justinkruit:
Want to do this before #814 ?

Issue #801, opened 2024-06-28
[BUG] Compile error using woocommerce css


justinkruit:
### Describe the bug

The following errors happen when trying to compile the Bootscore WooCommerce scss.

```
Warning: Array to string conversion in F:\LocalWP\Local Sites\bootscore\app\public\wp-content\themes\bootscore-main\inc\scssphp\src\Compiler.php on line 927

Warning: Array to string conversion in F:\LocalWP\Local Sites\bootscore\app\public\wp-content\themes\bootscore-main\inc\scssphp\src\Compiler.php on line 927

Warning: Array to string conversion in F:\LocalWP\Local Sites\bootscore\app\public\wp-content\themes\bootscore-main\inc\scssphp\src\Compiler.php on line 927
```

Seems to be related to this: https://github.com/scssphp/scssphp/issues/606

### Steps to reproduce

Compile the scss file while including `bootscore-woocommerce`.

crftwrk:
What is here exactly wrong? Compiles well and no errors?

justinkruit:
They're warnings which happen with each compile. According to the issue on scssphp it happens when a specific selector is being used.

crftwrk:
It's this selector:

https://github.com/bootscore/bootscore/blob/2382c5292568053c2538b36ea5487dc1605554a7/assets/scss/bootscore-woocommerce/_wc-qty-btn.scss#L27

cc @darrinbreen, @mswirydowicz

finalwebsites:
I get this warnings with every pageview and I have the idea that my site takes 5 seconds to load :(
![Screenshot 2024-09-02 om 12 28 10](https://github.com/user-attachments/assets/bc758546-9d61-4887-a921-3dd1262e47e5)

justinkruit:
@finalwebsites We are working on a fix (potentially #829). The higher loading times is because Bootstrap gets compiled on load. You can either turn off dev modus on WordPress, or add the following code to your child theme:
```
add_filter('bootscore/scss/skip_environment_check', '__return_true');
```

finalwebsites:
@justinkruit thanks a lot!

PR #823, opened 2024-08-12
Fix: Compile error using woocommerce css

crftwrk:
Closes https://github.com/bootscore/bootscore/issues/801

I solved this now using jQuery to get rid of the SCSS selector. It's not an elegant way, but it works and I think this is just temporary and we can revert this PR when scssphp v2 has fixed this. For this reason, the scss snippet is just commented out.

Go to https://dev.bootscore.me/product/beanie/, add product to the cart and check quantity input and buttons in offcanvas cart. Input and buttons are disabled, because this product has quantity `max="1"` attribute.

Edit: check https://github.com/bootscore/bootscore/pull/827

### DIFF
--- a/assets/scss/bootscore-woocommerce/_wc-qty-btn.scss
+++ b/assets/scss/bootscore-woocommerce/_wc-qty-btn.scss
@@ -22,19 +22,6 @@ WooCommerce Quantity Input
       margin: 0;
     }
   }
-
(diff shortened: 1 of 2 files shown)


justinkruit (approved):
I was hopeful #829 would work here, but it didn't. Seems like this is the only working solution for now

PR #827, opened 2024-08-24
Fix nested array comparison in matchExtends to prevent array-to-string conversion warnings

kemalkastrataj:
Fixed an issue where the use of array_diff on nested arrays caused PHP "Array to string conversion" warnings. We introduced a new method, arrays_are_equal_recursive, that recursively compares both flat and nested arrays, ensuring that valid comparisons are made without triggering errors. This ensures that even complex, nested array structures are handled correctly while maintaining the integrity of the comparison logic in matchExtends.

By replacing the flat array checks with a recursive comparison method, we resolved the issue without skipping valid comparisons, ensuring robustness and preventing potential logic gaps. This update follows best practices by keeping the array comparison logic modular and reusable, promoting maintainability and reducing the risk of future errors in complex array operations.

### DIFF
--- a/inc/scssphp/src/Compiler.php
+++ b/inc/scssphp/src/Compiler.php
@@ -895,6 +895,34 @@ class Compiler
     }
 
     /**
+     * Recursively compares two arrays, accounting for both flat and nested arrays.
+     *
+     * @param array $array1
+     * @param array $array2
+     * @return bool
+     */
+    protected function arrays_are_equal_recursive($array1, $array2)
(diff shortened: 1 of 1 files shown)


crftwrk:
May I ask why you closed this PR? It can fix https://github.com/bootscore/bootscore/issues/801 in a better way than https://github.com/bootscore/bootscore/pull/823. But maybe better to pr this to [scssphp](https://github.com/scssphp/scssphp).

PR #846, opened 2024-09-26
Footer 4 "widget" class assigned to wrong element

MattScarfe:
Moved the "widget" class from the `<h2> ` tag to the parent `<div>`  in Footer 4 widget area


crftwrk (approved):
Thank you!

PR #855, opened 2024-10-01
remove ajax cart from initial page load

Axos11:
Ajax cart was loaded in the html response as well as directly after the DOM is ready. So now the initial html repsonse should be cachable.

The "only" drawback is that to "prevent" flickering of the cart button (resizes when the total amount of the cart is set) I had to remove the display of the cart total from the cart button in the header.

I assume next time I should just delete the commented parts?

### DIFF
--- a/template-parts/header/actions-woocommerce.php
+++ b/template-parts/header/actions-woocommerce.php
@@ -61,12 +61,13 @@ if ( is_cart() ) {
     <div class="d-inline-flex align-items-center">
       <i class="fa-solid fa-bag-shopping"></i><span class="visually-hidden-focusable">Cart</span>
       <?php if (in_array('woocommerce/woocommerce.php', apply_filters('active_plugins', get_option('active_plugins')))) {
-        $count = WC()->cart->cart_contents_count;
+        //$count = WC()->cart->cart_contents_count;
         ?>
         <span class="cart-content">
+        <?php /**
           <?php if ($count > 0) { ?>
             <?= esc_html($count); ?>
-          <?php } ?>
+          <?php } ?> */ ?>
         </span>
       <?php } ?>
     </div>
--- a/template-parts/header/offcanvas-woocommerce.php
+++ b/template-parts/header/offcanvas-woocommerce.php
@@ -46,7 +46,7 @@ if ( is_checkout() || is_cart() ) {
     </div>
     <div class="offcanvas-body p-0">
       <div class="cart-list">
-        <div class="widget_shopping_cart_content"><?php woocommerce_mini_cart(); ?></div>
+        <div class="widget_shopping_cart_content"><?php //woocommerce_mini_cart(); ?></div>
       </div>
     </div>
   </div>
--- a/woocommerce/inc/ajax-cart.php
+++ b/woocommerce/inc/ajax-cart.php
@@ -406,7 +406,7 @@ function bootscore_ajax_add_to_cart_js() {
             }, 300);
 
             let cart_res = res.data;
-            $('.cart-content span.woocommerce-Price-amount.amount').html(
+            $('#offcanvas-cart .cart-content span.woocommerce-Price-amount.amount').html(
               cart_res['total']
             );
             wrap.find('.bootscore-custom-render-total').html(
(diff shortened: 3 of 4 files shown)


crftwrk:
Thanks for this PR, all in all it works well. There are two points:

Firstly, it will create issues with https://github.com/bootscore/bootscore/pull/841, because these snippets gets filters to change their classes, [view changes](https://github.com/bootscore/bootscore/pull/841/files#diff-bdb7bf75b471e06e8c79efe9741281b5138c18da8e5f54fb13b201f248af166e). So, we have to wait until https://github.com/bootscore/bootscore/pull/841 is merged.


Secondly,

> The "only" drawback is that to "prevent" flickering of the cart button (resizes when the total amount of the cart is set) I had to remove the display of the cart total from the cart button in the header.

The problem is that the amount in the shopping cart button will then be removed on thousands of sites. I'm sure that this will cause a lot of resentment among users and therefore I don't think it's worth it.

https://github.com/bootscore/bootscore/blob/8179d5d50540d4de32d4802cb6868a49fb3f5cb2/template-parts/header/actions-woocommerce.php#L63-L72

The question is if it's an optical thing that you don't like or if it has something to do with the functionality.

- If it is just an optical issue then price amount can easily be hidden by changing the classe to `d-none` by filters coming with https://github.com/bootscore/bootscore/pull/841.
- If this is related to the function, then I would suggest wrapping each snippet in a filter. This way the existing function would remain and users could "opt-in" to activate the new function with a filter, for example something like this:

```php
<?php
// Check if the filter is set to false (default behavior)
if (apply_filters('bootscore/remove_ajax_cart_from_initial_page_load', false) === false) {
  // WooCommerce cart logic
  if (in_array('woocommerce/woocommerce.php', apply_filters('active_plugins', get_option('active_plugins')))) {
    $count = WC()->cart->cart_contents_count;
    ?>
    <span class="cart-content">
      <?php if ($count > 0) { ?>
        <?= esc_html($count); ?>
      <?php } ?>
    </span>
    <?php
  }
} else {
  // Do nothing if filter is set to true
}
?>
```


`functions.php`:

```php
add_filter('bootscore/remove_ajax_cart_from_initial_page_load', '__return_true');
```


> I assume next time I should just delete the commented parts?

Yes, please ;-)

Any thoughts?

Axos11:
So the issue with the amount in the header is basically that the information isn't available at page load but e.g. 500ms later due to the ajax loading of the cart. that creates a resizing and potentially a [Cumulative Layout Shift (CLS)](https://web.dev/cls/) issue. Quite a small one because the "penalty" is relational to the size of the element but still.
And i really don't like resizing elements, I think its a bad user experience.

So, in my opinion it would definitely make sense to follow through with the changes, because risking the cart amount beeing cached makes no sense to me. 
I do understand that the change is potentially irritating to users of the theme, so i do like the idea of a filter.

Also, I think educating the user, that the existing approach slows down the page load and makes pages not cache-able would be best and maybe reverse the default behaviour in the next major version (e.g. 6.1 or 6.2). 

I don't like the "hidden" variant because then it still loads the information and slows down the initial page load with a query that is done via ajax a few ms after anyways.

To further add to the issue. On a plain bootscore/woocommerce install on a fast server the time difference is basically not measurable but as soon as some plugins are installed and the server is a little slower, the call to the mini cart function takes up significantly more time.

Ps: the filter would have to be applied also in the js parts to filter out the changes i did there.
I could push a suggestion including the filter if you like.

crftwrk:
Sorry for my late reply, pretty busy at the moment. The cart price will not roughly removed that way, but I'm sure that we will find a solution which fits to all needs and I think the filter is the best option. But please let's wait and revisit after https://github.com/bootscore/bootscore/pull/841 is merged, because there are changes in the same parts.

crftwrk:
@Axos11 https://github.com/bootscore/bootscore/pull/841 has been merged. Want to create a new PR with this filter?

Axos11:
OK, I can have another run at that. But I assume I can only start with that in 2 weeks or so as its quite busy right now.

crftwrk:
Replaced by https://github.com/bootscore/bootscore/pull/1032

PR #857, opened 2024-10-06
Adding Indonesian translation (100% translated)

shfrs:


crftwrk:
Thank you! Can you please move both files into the `languages` folder to get translation work?

shfrs:
I didn't realize I added it in the wrong folder, I've moved it now, thanks.

crftwrk (approved):
Thank you 🙏

PR #890, opened 2024-12-10
Enqueue main.css to WP Pattern Editor

crftwrk:
Adds missing compiled styles to the WP pattern editor to show patterns correctly **Backend > Appearance > Patterns**

| Before  | After |
| before  | after  |

### DIFF
--- a/inc/enqueue.php
+++ b/inc/enqueue.php
@@ -4,7 +4,7 @@
  * Enqueue styles & scripts
  *
  * @package Bootscore 
- * @version 6.0.0
+ * @version 6.0.3
  */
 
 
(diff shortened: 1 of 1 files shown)


inc/enqueue.php
```diff
@@ -51,22 +51,26 @@ function bootscore_scripts() {
...
+  add_action('admin_enqueue_scripts', function($hook_suffix) {
+    $screen = get_current_screen();
+    
+    // Enqueue editor.css only in the block editor
+    if ($screen && $screen->is_block_editor) wp_enqueue_style('editor-style', get_stylesheet_directory_uri() . '/assets/css/editor.css');
 
+    // Enqueue main.css only in the Pattern Library
+    if ('appearance_page_edit-wp-patterns' === $hook_suffix) wp_enqueue_style('bootscore-pattern-library-styles', get_stylesheet_directory_uri() . '/assets/css/main.css');
+  });
+}
```
justinkruit, line 72: Shouldn't this be outside the `after_setup_theme` action?

justinkruit (changes requested):

inc/enqueue.php
crftwrk: You're right, have splitted the editor and the pattern enqueues in two functions. See https://github.com/bootscore/bootscore/pull/890/commits/769ae8897aca42620a40bd350a3c631b19c38eeb

justinkruit (approved):

PR #909, opened 2024-12-25
Update scssphp 2.0.0 and revert #823

crftwrk:
- Revert https://github.com/bootscore/bootscore/pull/823, https://github.com/bootscore/bootscore/pull/909/commits/005d865d3fcb5db72690156d062f82aa98970ff1
- Update scssphp 2.0.0 https://github.com/bootscore/bootscore/pull/909/commits/9eb70a5e34c402c1a343555aacdd29eedcba508e, see https://github.com/scssphp/scssphp/releases/tag/v2.0.0

I have tested this and it seems to work fine from my point of view. But there are two downsides:

- scssphp drops support for PHP < 8.1, so PHP 8.1 is required
- scssphp is now quite heavy and increase the zip file up to 1,9 MB. I have checked for unneeded files, but everything is needed.


crftwrk:
@justinkruit let's do this later, think this needs more testing.

justinkruit:
@crftwrk alright 👍🏼

crftwrk:
I tested it, it works fine for me and I think we should merge it. @justinkruit, it would be great if you could double-check the overall functionality and merge it if you agree.

Issue #918, opened 2025-01-12
[BUG] Select2 not completely removed


MicheleGrimaldi:
### Describe the bug

Hi and thanks for this great theme! I'm working on a simple shop and I found out that in  Woocommerce > inc > wc-forms.php there's a snippet to dequeue the script "selectWoo". However it doesn't seem to work properly, cause in the checkout I still find the select2 input for picking up the province.

Searching online I found [this updated snippet](https://gist.github.com/ontiuk/72f6ac868d397678fb8b31df2b22e32a?permalink_comment_id=5320497#gistcomment-5320497) that works great! May you could update the theme file in the next release. What do you think?

crftwrk:
Hi,

> However it doesn't seem to work properly, cause in the checkout I still find the select2 input for picking up the province.

This script does not hide the select2, it disables the WC styles allowing to use the Bootstrap's `form-select` instead. Do you have a link to your project where you can show what do you mean?

MicheleGrimaldi:
The shop is still in development but I can give you an example showing a screenshot:
I'm using Bootscore with a child theme by the way.


Without editing the file I get this in the checkout: select2 with different height
<img width="991" alt="with-your-script" src="https://github.com/user-attachments/assets/1405a909-2202-4650-8fad-25edb1a59ff5" />

This is editing the file replacing the code I posted in the first comment:
<img width="1342" alt="with-my-script" src="https://github.com/user-attachments/assets/35250629-1046-47b7-a3ba-5ce2192c990d" />

This is weird, cause I checked on you website and it seems to work fine: I don't know if you have installed the latest version of Woo.

crftwrk:
>  I don't know if you have installed the latest version of Woo.

Of course, always using the latest versions.

<img width="1258" alt="Bildschirmfoto 2025-01-13 um 11 54 07" src="https://github.com/user-attachments/assets/9245ee47-f470-4688-ac6d-06ea63a41d9e" />

> The shop is still in development

Yes, but a screenshot is no help. It would be great if you can install two copies somewhere in the web and share the links for testing. One with actual snippet and one with the updated one. Then we can see, understand and fix it.

Thank you.

crftwrk:
Got it! You are using the blockified checkout instead the shortcode as described [here](https://bootscore.me/documentation/woocommerce/#cart-and-checkout-since-wc-8-3), right?

Agree that we can fix this quickly by changing the script to:

```php
/**
 * Remove CSS and/or JS for Select2
[...]

PR #919, opened 2025-01-14
Remove select2 scripts and styles on blockified classic checkout

MicheleGrimaldi:
Update snippet to replace woo Select2 even on cart and checkout blocks.

Closes https://github.com/bootscore/bootscore/issues/918


crftwrk (approved):
Perfect, thank you!

PR #926, opened 2025-01-17
Revert #919

crftwrk:
@MicheleGrimaldi unfortunately figured out that this creates a bug with Stripe which makes it impossible to pay by credit card. I quickly commented the code to revert your PR and reopened the issue. So we have to check and fix this later. Until this has been solved, please use shortcode in checkout instead the blockified classic cart as described in the [docs](https://bootscore.me/documentation/woocommerce/#cart-and-checkout-since-wc-8-3).

- https://github.com/bootscore/bootscore/issues/918
- https://github.com/bootscore/bootscore/pull/919

stripe

### DIFF
--- a/woocommerce/inc/wc-forms.php
+++ b/woocommerce/inc/wc-forms.php
@@ -24,7 +24,11 @@ function bootscore_dequeue_styles_and_scripts_select2() {
   wp_dequeue_script('selectWoo');
   wp_deregister_script('selectWoo');
 
-  // Dequeue and deregister the country select script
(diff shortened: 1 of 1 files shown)


massimoghm:
We can check if stripe plugin is in enabled? 
I think is difficult to debug if other plugins use WooCommerce Block select2 library...

crftwrk:
@massimoghm let's talk about this in the open issue https://github.com/bootscore/bootscore/issues/918#issuecomment-2623853165

Issue #934, opened 2025-01-24
[BUG] TinyMCE backend CSS on child theme


massimoghm:
### Describe the bug

Hi, 

I've tried on blank installation, there is a problem with TinyMCE CSS when scss build main.css on child.
If you tell me where to check I will try to solve it.

Best regards

### Steps to reproduce

1. Install Bootscore
2. Install Bootscore-Child
3. Install the plugin "Classic-Editor" 
4. Open a page and check the padding on TinyMCE, it is 0 for left and right. It change also Font-Family.

### Screenshots and Additional Info

Whit Bootscore
![Image](https://github.com/user-attachments/assets/d0b47725-5ca3-4367-8e0c-273c9f5a2220)

Whit Bootscore-Child enabled
![Image](https://github.com/user-attachments/assets/df01c551-5aa7-4113-927f-0d53f7a3d98a)

crftwrk:
I see, it gets overridden by the editor-style.css.  Adding this to your child's `_bootscore-custom.scss` will fix this roughly:

```css
body.mce-content-body {
  font-family: inherit;
  margin: 9px 10px;
}
```

If we want to fix this general in theme, I think we should do this in a better way.

Create new file `tinymce-editor.php` with following content and copy file into folder `inc`:

```php
<?php

/**
 * TinyMCE Editor
 *
 * @package Bootscore 
 * @version 6.0.5
 */


// Exit if accessed directly
defined('ABSPATH') || exit;

/*
 * Fix TinyMCE editor body margin and font-family that is overridden by Bootstrap editor-styles
 * https://github.com/bootscore/bootscore/issues/442
 * https://github.com/bootscore/bootscore/pull/757
 * https://github.com/bootscore/bootscore/issues/934
 */
add_filter( 'tiny_mce_before_init', function( $settings ) {

  $content_style = 'body.mce-content-body { margin: 9px 10px; font-family: inherit; }';

  if ( isset( $settings['content_style'] ) ) {
    $settings['content_style'] .= ' ' . $content_style;
  } else {
    $settings['content_style'] = $content_style;
  }

  return $settings;
});
```



Now link the file in `functions.php`:

```php
require_once('inc/tinymce-editor.php');          // Fix body margin and font-family in backend if classic editor is used
```

If you agree, a PR would be nice.

massimoghm:
You can find the fix in the pull for aria-label.
You need two different pull?

Best regards

crftwrk:
> You need two different pull?

[...]

PR #936, opened 2025-01-24
aria-label for hamburger menù

massimoghm:
aria-label for hamburger menù

Closes https://github.com/bootscore/bootscore/issues/934

### DIFF
--- a/header.php
+++ b/header.php
@@ -8,7 +8,7 @@
  * @link https://developer.wordpress.org/themes/basics/template-files/#template-partials
  *
  * @package Bootscore
- * @version 6.0.4
+ * @version 6.0.5
  */
 
 // Exit if accessed directly
--- a/header.php
+++ b/header.php
@@ -87,7 +87,7 @@ defined('ABSPATH') || exit;
           ?>
 
           <!-- Navbar Toggler -->
-          <button class="<?= apply_filters('bootscore/class/header/button', 'btn btn-outline-secondary', 'nav-toggler'); ?> <?= apply_filters('bootscore/class/header/navbar/toggler/breakpoint', 'd-lg-none'); ?> ms-1 ms-md-2 nav-toggler" type="button" data-bs-toggle="offcanvas" data-bs-target="#offcanvas-navbar" aria-controls="offcanvas-navbar" aria-label="<?php esc_html_e( 'Toggle main menu', 'bootscore' ); ?>>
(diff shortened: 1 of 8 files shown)


crftwrk:
Thank you, this is a good idea! 

If we do that, we should add the `aria-label`to all buttons in the header. This are the `search-toggler`, `account-toggler`, `mini-cart-toggler` and the `back-to-cart-toggler` located in `header-actions.php` and `header-actions-woocommerce.php` in folder `template-parts/header/`.

Want to add them?

massimoghm:
Updated, check if new escape method is correct. I think is more codex style.
Updated version to 6.1.0.

If you remember other accessibility points, tell me :)

Best regards

crftwrk:
Next release will be on Feb. 24 alongside the WooCommerce 9.7 update. Not sure if we are ready to ship v6.1 until there. So version 6.0.5 is more safe.

massimoghm:
Revert to 6.0.5 release :)

header.php
```diff
@@ -87,7 +87,7 @@
           ?>
 
           <!-- Navbar Toggler -->
-          <button class="<?= apply_filters('bootscore/class/header/button', 'btn btn-outline-secondary', 'nav-toggler'); ?> <?= apply_filters('bootscore/class/header/navbar/toggler/breakpoint', 'd-lg-none'); ?> ms-1 ms-md-2 nav-toggler" type="button" data-bs-toggle="offcanvas" data-bs-target="#offcanvas-navbar" aria-controls="offcanvas-navbar">
+          <button class="<?= apply_filters('bootscore/class/header/button', 'btn btn-outline-secondary', 'nav-toggler'); ?> <?= apply_filters('bootscore/class/header/navbar/toggler/breakpoint', 'd-lg-none'); ?> ms-1 ms-md-2 nav-toggler" type="button" data-bs-toggle="offcanvas" data-bs-target="#offcanvas-navbar" aria-controls="offcanvas-navbar" aria-label="<?php esc_html_e( 'Toggle main menu', 'bootscore' ); ?>>
```
crftwrk, line 90: Why using `esc_html_e`here while others uses `esc_attr_e`?

massimoghm: Changed

crftwrk (approved):
Thank you 🙏. Just fixed missing close `"` in [header.php](https://github.com/bootscore/bootscore/pull/936/commits/7ef8b91ca37244e1b257061db7116833f47dab84)

PR #947, opened 2025-02-10
Polish translation updates.

tools200ms:
I found that some polish translations is missing, so it has been added. Also entry:
```
msgid "Delivery details"
msgstr ""
"\n"
"Szczegóły dostawy"
```
was cause of an error: 
```
$ msgfmt -o pl_PL.mo pl_PL.po 
pl_PL.po:159: 'msgid' and 'msgstr' entries do not both begin with '\n'
msgfmt: found 1 fatal error
```
I fixed this part too.


crftwrk (approved):
Great, thank you 🙏

PR #965, opened 2025-02-26
Fix new Stripe checkout payment form

crftwrk:
- https://woocommerce.com/document/stripe/admin-experience/new-checkout-experience/
- https://woocommerce.com/document/stripe/customization/style-payment-form/
- https://docs.stripe.com/elements/appearance-api#supported-css-properties

Image


crftwrk:
Fixed by itself

Issue #974, opened 2025-03-03
TinyMCE in Backend keeps »growing« vertically in combination with wpbakery


typomaniac77:
### Describe the bug

I have the following issue together with wpbakery / js composer. In the backend editor, the Tiny MCE keeps ‘growing’ once it is opened until the window is filled and it keeps growing. with every click in the text field. This is not the case in other basic themes.

You can see the issue here:
https://www.youtube.com/watch?v=nxTkQ7aaA_c

crftwrk:
Thank you for pointing this out! It seems that this came with WP 6.7.2 which adds inline-styles to the editor iframe which is set to 100% height by Bootstrap. We can fix this by enqueuing the Bootstrap editor-style only if block editor is used. 

Go in parent theme to:


https://github.com/bootscore/bootscore/blob/e154b8f1a50104a1eb6c01d151f678d7082bafc0/inc/enqueue.php#L54-L62


And replace this code with:

```php
/**
 * Register editor styles for block editor.
 * Prevents loading when Classic Editor plugin is active.
 * https://github.com/bootscore/bootscore/issues/974
 */
function bootscore_add_editor_styles() {
  // Ensure function exists before calling it
  if (!function_exists('is_plugin_active')) {
    require_once ABSPATH . 'wp-admin/includes/plugin.php';
  }

  // Skip adding styles if Classic Editor is active
  if (is_plugin_active('classic-editor/classic-editor.php')) {
    return;
  }

  add_theme_support('editor-styles');
  add_editor_style('assets/css/main.css');
}
add_action('after_setup_theme', 'bootscore_add_editor_styles');
```

Let me know if that helps. If so, it would be great if you can fork the theme, fix that and create a pull request. If you do not know how to do this, I'm here to help.

typomaniac77:
Shouldn't it be:

  add_theme_support('editor-styles');
  add_editor_style('assets/css/editor.css'); 

instead of main.css?

typomaniac77:
ah! Okay! But the editor still resizes as long as main.css is added in line 71

faridulhassan:
@crftwrk , 
Please fix this issue asap. It's super annoying.

crftwrk:
@faridulhassan, Please follow this PR https://github.com/bootscore/bootscore/pull/975

PR #975, opened 2025-03-04
Update enqueue.php - fix for loading styles in classic editor

typomaniac77:
Closes https://github.com/bootscore/bootscore/issues/974

### DIFF
--- a/inc/enqueue.php
+++ b/inc/enqueue.php
@@ -52,15 +52,25 @@ add_action('wp_enqueue_scripts', 'bootscore_scripts');
 
 
 /**
- * Register editor styles.
+ * Register editor styles for block editor.
+ * Prevents loading when Classic Editor plugin is active.
+ * https://github.com/bootscore/bootscore/issues/974
  */
 function bootscore_add_editor_styles() {
-  // Add support for editor styles and main.css for the editor
+  // Ensure function exists before calling it
+  if (!function_exists('is_plugin_active')) {
+    require_once ABSPATH . 'wp-admin/includes/plugin.php';
+  }
+
+  // Skip adding styles if Classic Editor is active
+  if (is_plugin_active('classic-editor/classic-editor.php')) {
+    return;
+  }
+
   add_theme_support('editor-styles');
   add_editor_style('assets/css/main.css');
 }
 add_action('after_setup_theme', 'bootscore_add_editor_styles');
-
 
 /**
  * Enqueue styles for block editor and Pattern Library.
--- a/inc/enqueue.php
+++ b/inc/enqueue.php
@@ -4,7 +4,7 @@
  * Enqueue styles & scripts
  *
  * @package Bootscore 
- * @version 6.0.3
+ * @version 6.1.1
  */
 
 


crftwrk:
It seems that this needs more attention. https://github.com/bootscore/bootscore/pull/976 does not enqueue editor-style if Block editor is in use, so I closed it and let's go further here.

Have following improved idea which enqueues editor-styles only if Block editor is in use. It works at first glance fine with Block, Classic and in general if TinyMCE is active:

```php
/**
 * Register editor styles only if the block editor is active.
 */
function bootscore_add_editor_styles() {
  add_action('current_screen', function($screen) {
    if (function_exists('use_block_editor_for_post_type') && use_block_editor_for_post_type($screen->post_type)) {
      add_theme_support('editor-styles');
      add_editor_style('assets/css/main.css');
    }
  });
}
add_action('after_setup_theme', 'bootscore_add_editor_styles');
```

Let me know what you think!

typomaniac77:
great! That works!

crftwrk:
Unfortunately, I discovered one more issue. If editor-style is enqueued like in my comment abofy, styles are not added to the pattern library Backend > Appearance > Patterns > Bootscore. Means we have to refactor the entire scripts:

https://github.com/bootscore/bootscore/blob/d8338834f8b94e363b9c36cdcb44799f275aa1bb/inc/enqueue.php#L54-L81

crftwrk:
@typomaniac77 Let's try another approach. 

- Restore the default `enqueue.php`, no changes here https://github.com/bootscore/bootscore/blob/d8338834f8b94e363b9c36cdcb44799f275aa1bb/inc/enqueue.php
- Adding inline CSS to TinyMCE instead is a simple solution.

Go to:

https://github.com/bootscore/bootscore/blob/d8338834f8b94e363b9c36cdcb44799f275aa1bb/inc/tinymce-editor.php#L22

and add `height: auto;`:

```php
  $content_style = 'body.mce-content-body { margin: 9px 10px; font-family: inherit; height: auto;}';
```

This works fine in my case for block and classic editor. I tried to do this changes by myself in your branch but I have no access to it.

@typomaniac77 @faridulhassan This have to be deeply tested in both, **block and classic editor** on your side as well. If you agree with this changes, we can fix this bug quickly.

faridulhassan:
@crftwrk , 
Though the tinymce height is fixed in this way, but child theme's main.css file is loading into dashboard editing page (Screenshot attached). 

Your following approach worked fine for me. It fixes the auto height issue and doesn't load main.css in dashboard page.

`function bootscore_add_editor_styles() {
  add_action('current_screen', function($screen) {
    if (function_exists('use_block_editor_for_post_type') && use_block_editor_for_post_type($screen->post_type)) {
      add_theme_support('editor-styles');
      add_editor_style('assets/css/main.css');
    }
  });
}
add_action('after_setup_theme', 'bootscore_add_editor_styles');`

crftwrk:
@faridulhassan 

> Though the tinymce height is fixed in this way

Great, are you able to create quickly a new pull request? Then we can merge this directly and release a new patch soon.

> but child theme's main.css file is loading into dashboard editing page 

Yes, that's how WordPress handles it and it's intended. You can check this by installing a default WP theme https://wordpress.org/themes/twentytwentyone/ and check the Classic Editor (Styles) in the backend.

But we can wrap the editor-styles enqueue in a filter in a second pull request to disable editor-styles with a single line in child's functions.php. For example like this:


```php
/**
 * Disable editor-style
 */
add_filter('bootscore/load/editor_styles', '__return_false');
```

Do you agree?

crftwrk:
@faridulhassan The more I think about it, I believe this is a bug (or it is intended to move users to the GB editor) came with WordPress 6.7.2. It does not happen on earlier WP versions and default themes editor-style is enqueued to classic editor as well. 

There is no way to disable editor-styles in classic editor, we can only stop the growing by adding the height CSS to TinyMCE.

crftwrk:
https://github.com/bootscore/bootscore/pull/983

PR #976, opened 2025-03-04
Update enqueue.php > change assigned styles for classic editor

typomaniac77:


crftwrk:
check: https://github.com/bootscore/bootscore/issues/974#issuecomment-2697205626

crftwrk:
This does not enqueue the editor-styles if GB Block editor is in use. It has to work for both, Block and TinyMCE editor. Let's go in more detail here https://github.com/bootscore/bootscore/pull/975

PR #981, opened 2025-03-10
Updated french translation for "Open side menu"

fnicollet:
Set it to "Ouvrir le menu latéral" instead of the default english label "Open side menu":
image


crftwrk (approved):
Thank you! 👍

PR #990, opened 2025-03-20
Fixed typo in fr_FR.po

Foalyy:


Foalyy:
Sorry for the double commit, I naively modified the .po file at first without compiling it. I properly edited it using Poedit in the second commit.

crftwrk (approved):
Merci 🙏

PR #1010, opened 2025-04-19
Feature/ajax login

Axos11:
* Creates an ajax login for woocommerce in the offcanvas user account.
 *
 * The implementation has 2 main goals
 * 1. Better user experience. Login happens immediately and the user can directly access his account
 * 2. Removes one of the last parts of the template that makes full page caching possible even for logged in users.
 *
 * Optout of ajax login: add_filter('bootscore/wc_ajax_login', '__return_false');

#### Docs

- [x] Filters and action hooks - Republish draft
- [x] Migration - Republish draft
- [ ] Translations
- [x] WooCommerce - Republish draft
   - [x] Sensei


crftwrk:
@Axos11 @MarkusKab Thank you, this is hot!

We need to do some changes before merging. Tried to do them directly in this PR but I have no permission to change files. So, I created a PR to your branch https://github.com/risingBYTES/bootscore/pull/1. If you agree, please merge it, then your PR will be updated.

Check https://dev.bootscore.me, just type something into the login form and click the login button. This PR does following:

- Deleted unwanted language backup files that have been created by Loco Translate. This removes 6500 lines of unused code ;-)
- Removes the last empty line in `functions.php`. Think this line was mistakenly added, but there are no changes in this file. Just to avoid confusion and probably conflicts with other PR's.
- Added missing file headers and updated existing file versions.
- Formatted all files from TAB to 2 spaces like all other files in the theme.

### Loader

If you have a slow connection, then login offcanvas opens empty and it need some seconds to show the form. When you click the Login button, it seems that nothing happens. But we have to tell the user that something happens when offcanvas is open and when clicked the Login button.

- Added a class `ajax-loading` to the body in `ajax-login.js` This class is added while something loads and removed when loaded.
- Added an action `<?php do_action( 'bootscore_before_my_offcanvas_account' ); ?>` to `offcanvas-woocommerce.php`.
- Hooked a loader to this action in `wc-ajax-login.php`

Now, when open the offcanvas or clicking the login button, a loader is shown to tell the user that there is something in progress.

Please review the files and if you agree, merge it into your branch. When done, we should test this on some live sites for one week and share links to them before merging your PR into the theme.

#### Just for the docs

New Action:

```php
<?php do_action( 'bootscore_before_my_offcanvas_account' ); ?>
```

New Filter:

```php
/*
 * Disable AJAX login
 */
add_filter('bootscore/wc_ajax_login', '__return_false');
``` 

Agree with this changes?

crftwrk:
@Axos11 thanks for merging. Have installed this here for testing:

- https://bootscore.me
- https://www.vdzev.de/
- https://wilob.ch/

Let's test this for at least one week before merging in to the theme. 

- Triple check the translation files, maybe you're right and we have to re-add them.
- Maybe a better solution for the `body` loader class?

If you have this installed on live sites as well, please share links.

crftwrk (approved):
👍

PR #1051, opened 2025-06-24
Improves IP address detection

Axos11:
Enhances IP address retrieval to accommodate hosts that do not populate standard $_SERVER variables correctly.

This ensures reliable IP detection across a wider range of server configurations.

Discussion: https://github.com/orgs/bootscore/discussions/1049,


Axos11:
@crftwrk We could think of putting the IP detection behind a filter. So that people could either disable it if wanted or write their own custom logic if needed.

As of now, I assume that should fix it for most people.

crftwrk (approved):
🙏

Issue #1060, opened 2025-07-16
[Feature Request] accessibility of the to_top_button


neralex:
### Describe the feature that you'd like

While fixing some accessibility issues based on the WCAG 2.1 rules, I just stumbled across this pseudo-button. From a technical perspective, it doesn't really make sense to use an anchor-tag, since this element mimics the functionality of a button, but it doesn't have an associated role-attribute, which wouldn't be necessary if the element is a vanilla HTML button tag. And if aria-label is already added to the element, then the nested span.visually-hidden-focusable is no longer necessary because aria-label already does the job.

So I had to override the footer.php from the main-theme to the child theme to change it to:

`<button class="<?= apply_filters('bootscore/class/footer/to_top_button', 'btn btn-primary shadow'); ?> position-fixed zi-1000 top-button" aria-label="<?php esc_attr_e('Return to top', 'bootscore' ); ?>"><?= apply_filters('bootscore/icon/chevron-up', '<i class="fa-solid fa-chevron-up" aria-hidden="true"></i>'); ?></button>`

But then the scroll-functionality doesn't work anymore and I had to add a extra click-function for this simple button. 

1. Where is the scroll-functionality in the main-theme placed for this pseudo-button?
2. Is there chance to change this pseudo-button to an vanilla HTML button without the nested span.visually-hidden-focusable to avoid override the whole footer.php from the main theme?

Thanks in advance

### Checklist

- [x] I have searched the issues and this feature has not been requested.

crftwrk:
> Where is the scroll-functionality in the main-theme placed for this pseudo-button?

There is no scroll-functionality in the theme , the to-top button uses a simple `href="#"` which scrolls to the top of the page and scrolling effect is made by default Bootstrap:

```css
html {
  scroll-behavior: smooth;
}
```


This is a simple and modern way to add a to-top button as written by [CSS-Tricks](https://css-tricks.com/how-to-make-an-unobtrusive-scroll-to-top-button/#aa-and-what-about-smooth-scrolling).

In version 4, we had a button with JS but luckily since all modern browsers supports `scroll-behavior` there is no need for this anymore.

If you want to change the `<a>` to a `<button>` element, you have to add additional JS.

The only JS still exists is to hide/show the button on scroll position: 

https://github.com/bootscore/bootscore/blob/a0950beef48bd3bf2b1413519a6bd2ddba41720b/assets/js/theme.js#L23-L32


[...]

PR #1068, opened 2025-07-25
Improve accessibility of the to top button

crftwrk:
- Added role="button" for screen reader compatibility
- Removed visually-hidden span and used aria-label for translation support

Closes https://github.com/bootscore/bootscore/issues/1060

@neralex Do you agree with this changes?

### DIFF
--- a/footer.php
+++ b/footer.php
@@ -8,7 +8,7 @@
  * @link https://developer.wordpress.org/themes/basics/template-files/#template-partials
  *
  * @package Bootscore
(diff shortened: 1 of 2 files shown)


neralex:
> * Added role="button" for screen reader compatibility
> 
>     * Removed visually-hidden span and used aria-label for translation support
> 
> 
> Closes #1060
> 
> @neralex Do you agree with this changes?

Yeah sure!

PR #1094, opened 2025-09-01
linking parents, current category as text

neralex:
When you are reaching the archive of a category, the link of the category was still shown at the breadcrumb-navigation because there is no check existing, if you are already inside of a specific categories-archive. For this case, the link to the same category isn't needed at the breadcrumb-navigation because there is no other target available, than the categoy itself.


crftwrk (approved):
This is a good one, thank you!

PR #1103, opened 2025-09-11
PHP 8 fix to avoid this warning: Undefined array key "input_value"

neralex:
Problem: $args[‘input_value’], min_value, max_value are not guaranteed to be assigned > PHP 8 throws notices.

_Warning: Undefined array key "input_value" in \wp-content\themes\bootscore\woocommerce\inc\wc-qty-btn.php on line 31_

The issue happens under php 8.3.2x while trying to reach the checkout.

The added safeguards ensure that meaningful, complete, type-consistent values are always available - regardless of what Woo/plugins deliver - eliminating PHP 8 notices and keeping button states correct.

It works for me. But please test the changes to see if the snippet still works for the cases you have added it for.

### DIFF
--- a/woocommerce/inc/wc-qty-btn.php
+++ b/woocommerce/inc/wc-qty-btn.php
@@ -4,7 +4,7 @@
    * WooCommerce Quantity Buttons
    *
    * @package Bootscore
-   * @version 6.2.0
+   * @version 6.3.1
    */
 
 // Exit if accessed directly


crftwrk (approved):
Thank you, I'm in 👍

@Axos11 can you please double check?

Edit: https://github.com/orgs/bootscore/discussions/1105

Axos11:
The question is "why" is woocommerce loading that file in the checkout in the first place. Because in our templating we don't have any qty buttons on the checkout page. So basically I think both should be cleaned up (prevent loading, find out what actually loads the file in the first place) and the sanitation added as well.
I have to double check the 0 stuff. But otherwise code looks fine to me. Will try to do that asap

Axos11:
@neralex every contribution is welcome and thanks for bringing the focus back to that issue. I think that I would not change any of the values without need but just escape early if the data needed for the logic isn't present.

@crftwrk what do you think?

```php
  add_filter('woocommerce_quantity_input_args', function (array $args) {

    add_filter('woocommerce_quantity_input_type', 'change_quantity_input_type');

    // Exit early if required args are not set
    if (!isset($args['max_value']) || !isset($args['min_value']) || !isset($args['input_value'])) {
      return $args;
    }
    // Rest of the code
```

neralex:
@Axos11 I think it's great because 1. it fixes the bug and 2. it changes the actual behaviour less invasively than my patch does. #shortandsimple

neralex:
@crftwrk thanks for the offer but just merge it. @Axos11 found a great solution.

crftwrk:
Got it!

PR #1108, opened 2025-09-26
Switch back to JS checkout validation

crftwrk:
WIP!

- Reverts https://github.com/bootscore/bootscore/pull/563


crftwrk:
This is a complete mess, we leave it as it is.

PR #1133, opened 2025-11-27
Check for missing escape

massimoghm:
I've checked the code to add missing escaping where the Codex suggests its use.
Please, if you think it’s useful, check all changes before merging, because I might make mistakes in some places where HTML can be used.
Regards!

### DIFF
--- a/inc/blocks/disable-unsupported-blocks.php
+++ b/inc/blocks/disable-unsupported-blocks.php
@@ -1,10 +1,10 @@
 <?php
 
 /**
- * Experimental: Disable unsupported blocks and patterns - Whitelist
+ * Disable unsupported blocks and patterns - Whitelist
  *
  * @package Bootscore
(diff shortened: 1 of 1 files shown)


crftwrk:
@massimoghm,

i pushed also a commit that uses `Template Version: 6.x.x` instead `@version 6.x.x` for template files.

- It's the default WP codex
- Let's quick view if that is a template or a function file
- Allows us to add a script that check if there are outdated templates in child and shows a warning in the backend

I have deeply checked and found no issues, great work! But impossible to check every filter, so maybe we will find some issues later. 

I have sent you 2 child-themes via WeTransfer, which have a heavy usage of almost all filters. If they have no issues, this PR should be fine.

I am fine here now. If you are fine as well, we can merge this.

Thank you!

crftwrk (dismissed):
🙏

PR #1209, opened 2026-05-27
apply filters for the subtotal calculation for better compatibility + extend BFcache fix to all browsers

Axos11:
In most cases there was no issue not having that filter applied, but with plugins relying on that filter issues came up. Especially if the same item existed twice in the cart with different subtotals, then both items would get the same subtotal applied. Only visually though. The calculation in the background was fine all the time.


crftwrk (approved):
👍

PR #1140, opened 2025-11-29
Feature refactor wc product filter block integration

crftwrk:
@murilocarvalhodev,
    
created a new branch https://github.com/bootscore/bootscore/tree/Feature-Refactor-WC-product-filter-block-integration and merged your PR into it. New PR is now https://github.com/bootscore/bootscore/pull/1140.
    
Reason for that is that there is a lot of movement in the main branch right now and its hard to keep your PR branch up-to-date. So, your commits are save. Please update your fork to the latest Bootscore main and submit changes from a new and updated branch to the new PR branch.
    
Demo: https://dev.bootscore.me/shop/ (Sidebar, works only on shop pages)

screenshot

### Issues

- [x] Active filters removes the add-to-cart `btn-loader` https://github.com/bootscore/bootscore/pull/1210
- [x] Active filters removes the cart toast alert in loop https://github.com/bootscore/bootscore/pull/1212
- [x] Check range in Firefox https://github.com/bootscore/bootscore/pull/1140/changes/70e3ec3e3cec5bd6ab88a9710cfff2af0d754d3d
- [x] ~~Disable the range forms? Changing the value there does nothing~~
- [x] ~~Add filters to clear button and badges~~
- [x] ~~Move more SCSS stuff to classes~~

### New Filters

```php
bootscore/block/product/filters/content
```

PR #1210, opened 2026-05-28
btn-loader fix for multiple elements in the query string

Axos11:
Fix the str_replace on the archive loop to also work if add-to-cart is not the first part of the query string


crftwrk (approved):
👍 This is exactly what we need for https://github.com/bootscore/bootscore/pull/1140. Thank you, please merge!

PR #1212, opened 2026-05-29
refactor ajax add to cart functionality to a function to be applied on dynamically created content

Axos11:


crftwrk (approved):
That's it! 👍🙏

PR #1225, opened 2026-07-03
added function_exists-check for bootscore_breadcrumb_shortcode()

neralex:
The shortcode-function inside of the breadcrump.php is missing a functions_exists-check to prevent the php error "Cannot redeclare bootscore_breadcrumb_shortcode()" after overriding this file inside of the child-theme.


crftwrk (approved):
🙏

PR #1245, opened 2026-08-27
v7: Refactor site logo to `picture` tag and add fixed width & height …

crftwrk:
…plus a filter

https://stackoverflow.com/questions/74638826/alternate-images-based-on-users-color-preference-without-using-javascript


crftwrk:
This does not work with `data-bs-theme="dark"`, nice try...

PR #1249, opened 2026-08-30
Namespace submit handler, scope wp_loaded removal to AJAX requests

### LINKED ISSUE #1234: [BUG] Bootscore AJAX add-to-cart: third-party form plugins lose their POST data
### Describe the bug

**Status:** bug report + upstream API request
**Reported by:** Claude code - dev site (bootscore-child on WooCommerce)
**Date:** 2026-08-05
**Affects:** `bootscore/woocommerce/js/ajax-cart.js` + `bootscore/woocommerce/inc/ajax-cart.php`
(observed on Bootscore 6.4.0 file headers; parent theme reported as 6.3.1)

---

## 1. Summary

Bootscore's single-product AJAX add-to-cart handler captures the form payload as
the **first statement of its submit handler**. Any plugin that populates hidden
inputs *during* the submit event — which is the standard pattern for form plugins
that need to describe client-side state to the server — has its fields silently
discarded.

We hit this with **WS Form PRO + WS Form WooCommerce**, where it caused every
custom product field to vanish from the cart item. The failure is completely
silent: no PHP error, no validation notice, and a clean success response.

We believe this affects any plugin using the same pattern, not just WS Form, and
that a small change in Bootscore would make the whole class of conflict tractable.

## 2. Symptom

Configurable product with seven conditionally-shown `select` fields. Client side
verified healthy: conditional logic resolves correctly, and the intercepted
request payload **does** contain `field_51[] = 24 TPI — Standard Raker`.

Server side, none of it lands:

- Repeated adds with different configurations **merge into one line item**, so the
  cart item data hash is identical — the custom meta is definitively absent.
- The endpoint returns a normal fragments + `cart_hash` response. No error, no
  notice. Nothing signals a problem.
- Removing Bootscore's submit handler to force a native POST adds **nothing at
  all** to the cart (see §5).

Notably, the misdirection here is expensive: the obvious hypothesis is "the theme's
custom endpoint bypasses `WC_Form_Handler`". It does not. That sent us down the
wrong path for a while, and we expect it will do the same to others.

## 3. Root cause

`?wc-ajax=ace_add_to_cart` **does** route through core:

```php
// bootscore/woocommerce/inc/ajax-cart.php:90
function bootscore_ajax_add_to_cart_handler() {
  WC_Form_Handler::add_to_cart_action();
  WC_AJAX::get_refreshed_fragments();
}
```

So `woocommerce_add_cart_item_data` fires and WS Form's filter runs. The data is
lost one step earlier — in the browser.

WS Form injects its control fields (`wsf_form_id`, `wsf_hash`, `wsf_hidden`,
`wsf_bypass_required`, nonce) **only at submit time**, inside `form_post()`
(`ws-form-pro/public/js/ws-form-public.js:3018-3115`). None of them exist in the
DOM until the user clicks. Meanwhile:

```js
// bootscore/woocommerce/js/ajax-cart.js:19-32
$('form.cart').on('submit', function (e) {
  e.preventDefault();
  const form = $(this);
  var button = form.find('.single_add_to_cart_button');
  if (button.hasClass('disabled')) { return; }
  var formData = new FormData(form[0]);   // ← captured before WS Form writes anything
```

Both handlers are bound to the same `form.cart` element — WS Form resolves its
form object as `form_canvas_obj.closest('form')`
(`ws-form-pro/shared/js/ws-form.js:369`) and binds submit on it
(`ws-form-public.js:2211`). Bootscore binds at document-ready and, in practice,
binds first; jQuery runs same-element handlers in binding order, and
`preventDefault()` does not stop the second handler. So Bootscore posts a stale
snapshot, and WS Form's server-side filter bails on `0 === $form_id`.

WS Form then completes its own pipeline and calls `ws_form.form_obj[0].submit()`
(`ws-form-woocommerce/js/public.js:298`) — a native submit that `preventDefault()`
cannot suppress. We observed this as an occasional **duplicate add**.

**The ordering is load-order dependent, which makes this worse, not better.** If
WS Form happened to bind first its fields would be present — so the bug can appear
and disappear based on enqueue order, caching, or optimisation plugins.

## 4. Reproduction

1. WooCommerce → Settings → Products → enable AJAX add-to-cart buttons.
2. Install WS Form PRO + WS Form WooCommerce; attach a form with any saved field
  to a product.
3. Add the product to the cart from its single product page.
4. Inspect the cart item: no WS Form meta. Add again with a different field value:
  it merges into the same line item instead of creating a second one.

## 5. Why the obvious workarounds don't work

**"Dequeue `ajax-cart.js` on those product pages."** Two problems:

1. `ajax-cart.php:99` removes the core handler **globally and unconditionally**:
  
  ```php
  remove_action('wp_loaded', array('WC_Form_Handler', 'add_to_cart_action'), 20);
  ```
  
  That is PHP-side, so dequeuing JS doesn't restore it. A native form POST then
  adds nothing to the cart. There is no working fallback path to fall back *to*.
2. `ajax-cart.js` is not only add-to-cart. It also owns the mini-cart quantity
  +/− buttons, remove-from-cart, the offcanvas toasts and the BFCache refresh
  (lines 151-407). Dropping the file breaks the offcanvas cart on that page.

**"Use the `bootscore/load_ajax_cart` filter."** It's all-or-nothing site-wide and
evaluated on `after_setup_theme`, long before the queried product is known.

## 6. Our workaround

Child theme, ~90 lines, no parent or plugin files touched:

1. Unbind Bootscore's `form.cart` submit handler.
2. No-op `formEl.submit` so WS Form's native submit can't reload the page.
3. Re-issue the same `ace_add_to_cart` request from WS Form's
  `wsf-submit-complete` event, inside `setTimeout(…, 0)` so the WooCommerce
  add-on's own handler stamps its hidden inputs first.

This works and fixed the duplicate add. **But step 1 is genuinely fragile**, and
that fragility is the reason for this brief. With no namespace on the handler and
no opt-out, the only way to remove Bootscore's handler without also removing
WS Form's is to identify it by its function body:

```js
var events = $._data($form[0], 'events');
$.each(events.submit.slice(), function (i, h) {
  if (String(h.handler).indexOf('ace_add_to_cart') !== -1) {
    $form.off('submit', h.handler);
  }
});
```

That depends on a **private jQuery API** (`$._data`) and on the **literal string
`ace_add_to_cart` appearing inside the handler**. Rename the endpoint and this
silently stops matching — the handler stays bound, and the original data-loss bug
returns with no error anywhere.

## 7. Requests

In rough order of value per unit of effort:

**(a) Namespace the submit handler.** One line, no behaviour change, removes the
need for string matching entirely:

```js
$('form.cart').on('submit.bootscoreAjaxAddToCart', function (e) { … });
```

There's already precedent — the loop handler is namespaced
`click.bootscoreAjaxAddToCart` (`ajax-cart.js:88`). Third parties could then do a
supported `$('form.cart').off('submit.bootscoreAjaxAddToCart')`.

**(b) Give `form.cart` a way to opt out.** The handler already special-cases
`.product-type-external`; generalising that to a data attribute would cover this
whole class of conflict:

```js
if ($(this).is('[data-bs-no-ajax-cart]')) { return; }   // let the form submit natively
```

paired with a PHP filter, e.g.
`apply_filters('bootscore/woocommerce/ajax-cart/single/enabled', true, $product)`,
that emits the attribute. Integrators could then opt a product out declaratively
instead of unbinding handlers.

**(c) Scope the `wp_loaded` removal to actual AJAX requests.** As written,
`ajax-cart.php:99` disables the native add-to-cart path site-wide, so opting out
of the AJAX handler leaves nothing working. `WC_AJAX::do_wc_ajax` runs on
`template_redirect` (priority 0) while `add_to_cart_action` runs on `wp_loaded`
(priority 20), so the removal only needs to apply when the AJAX endpoint is the
one being served:

```php
add_action('wp_loaded', function () {
  if (isset($_GET['wc-ajax']) && 'ace_add_to_cart' === $_GET['wc-ajax']) {
    remove_action('wp_loaded', array('WC_Form_Handler', 'add_to_cart_action'), 20);
  }
}, 19);
```

That keeps double-add protection on the AJAX route while leaving a working native
fallback for everything else.

**(d) Emit a cancelable event before the request.** Triggering something like
`bootscore_ajax_add_to_cart_before` on `document.body` with the form, and bailing
if a handler returns `false`, would let plugins defer or take over the request
without touching the binding at all.

**(e) At minimum — and this is the ask we'd most like honoured even if (a)–(d) are
declined:** if the `ace_add_to_cart` endpoint name, the handler binding, or the
FormData capture point changes, please **call it out in the release notes** and
document that integrations may be matching on the endpoint string. Silent renames
here break downstream integrations in a way that produces no error message and no
failed request — just quietly missing cart data. A one-line changelog entry would
save someone the multi-hour trace this took us.

## 8. Offer

Happy to open a PR for (a) and (c) — both are small and self-contained — and to
test any of the above against this WS Form setup, which reproduces the bug
reliably. Also happy to supply our child-theme bridge as a reference
implementation if it's useful as a compatibility note for form-plugin users.

---

### Reference — file:line index

| File | Line | Relevance |
| --- | --- | --- |
| `bootscore/woocommerce/js/ajax-cart.js` | 19-32 | Submit handler; FormData captured before third-party fields exist |
| `bootscore/woocommerce/js/ajax-cart.js` | 88  | Loop handler — already namespaced (precedent for request (a)) |
| `bootscore/woocommerce/js/ajax-cart.js` | 151-407 | Mini-cart qty / remove / toasts / BFCache — why dequeuing is not viable |
| `bootscore/woocommerce/inc/ajax-cart.php` | 90-96 | Endpoint *does* call `WC_Form_Handler::add_to_cart_action()` |
| `bootscore/woocommerce/inc/ajax-cart.php` | 99  | Unconditional global removal of the core `wp_loaded` handler |
| `bootscore/woocommerce/wc-functions.php` | 45-53 | `bootscore/load_ajax_cart` — site-wide, too early to be product-aware |
| `ws-form-pro/public/js/ws-form-public.js` | 3018-3115 | WS Form injects its hidden inputs at submit time |
| `ws-form-pro/public/js/ws-form-public.js` | 2211 | WS Form binds submit on the same `form.cart` |
| `ws-form-pro/shared/js/ws-form.js` | 369 | `form_obj = form_canvas_obj.closest('form')` |
| `ws-form-woocommerce/js/public.js` | 298 | Native `form_obj[0].submit()` — source of the duplicate add |
| `ws-form-woocommerce/includes/classes/class-ws-form-action-woocommerce.php` | 2449 | `if (0 === $form_id) return;` — where the data is dropped |

### Steps to reproduce

_No response_

### Screenshots and Additional Info

_No response_

### Website link

_No response_

seanvandermolen:
Addresses requests **(a)** and **(c)** of #1234.

### (a) Namespace the single-product submit handler

`woocommerce/js/ajax-cart.js` now binds `submit.bootscoreAjaxAddToCart` instead of a bare `submit`, matching the precedent already set by the loop handler's `click.bootscoreAjaxAddToCart`.

Third parties can unbind it in a supported way:

```js
$('form.cart').off('submit.bootscoreAjaxAddToCart');
```

instead of walking the private `$._data()` API and string-matching `ace_add_to_cart` inside the handler body, which silently stops matching if the endpoint is ever renamed.

No behaviour change.

### (c) Scope the `wp_loaded` removal to actual AJAX requests

`woocommerce/inc/ajax-cart.php` removed `WC_Form_Handler::add_to_cart_action` globally and unconditionally, which also disabled the native (non-AJAX) form POST site-wide — so opting out of the JS handler left nothing working to fall back to.

The removal is now deferred into a `wp_loaded` callback at priority 19 that only fires when `ace_add_to_cart` is the endpoint being served. `WC_AJAX::do_wc_ajax()` runs on `template_redirect` (priority 0) while `add_to_cart_action` runs on `wp_loaded` (priority 20), so the requested endpoint is known in time.

Double-add protection on the AJAX route is unchanged; every other request keeps a working native form POST.

The priority-19 hook registers in time because `ajax-cart.php` is required on `after_setup_theme`. `WC_AJAX::get_endpoint()` always builds the URL as a `wc-ajax` query arg, so the `$_GET` check is reliable — there is no rewrite-based path to the endpoint.

### Not included

Requests (b) opt-out attribute, (d) cancelable event and (e) documentation are not part of this PR. Happy to follow up on any of them.

Closes https://github.com/bootscore/bootscore/issues/1234

### DIFF
--- a/woocommerce/inc/ajax-cart.php
+++ b/woocommerce/inc/ajax-cart.php
@@ -95,8 +95,26 @@
     add_action('wc_ajax_ace_add_to_cart', 'bootscore_ajax_add_to_cart_handler');
     add_action('wc_ajax_nopriv_ace_add_to_cart', 'bootscore_ajax_add_to_cart_handler');
 
-    // Remove WC Core add to cart handler to prevent double-add
-    remove_action('wp_loaded', array('WC_Form_Handler', 'add_to_cart_action'), 20);
+    /**
+     * Remove the WooCommerce core add to cart handler to prevent a double-add,
+     * but only while our own AJAX endpoint is the one being served.
+     *
+     * WC_AJAX::do_wc_ajax() runs on template_redirect (priority 0), while
+     * WC_Form_Handler::add_to_cart_action() runs on wp_loaded (priority 20), so the
+     * decision can be deferred until the requested endpoint is known. Removing it
+     * unconditionally would also disable the native (non-AJAX) form POST, leaving no
+     * working fallback for forms that opt out of the JS handler.
+     */
+    function bootscore_maybe_remove_wc_add_to_cart_action() {
+      // phpcs:ignore WordPress.Security.NonceVerification.Recommended -- Only reads the endpoint name, no data is processed here.
+      $wc_ajax_endpoint = isset($_GET['wc-ajax']) ? sanitize_text_field(wp_unslash($_GET['wc-ajax'])) : '';
+
+      if ('ace_add_to_cart' === $wc_ajax_endpoint) {
+        remove_action('wp_loaded', array('WC_Form_Handler', 'add_to_cart_action'), 20);
+      }
+    }
+
+    add_action('wp_loaded', 'bootscore_maybe_remove_wc_add_to_cart_action', 19);
 
 
     /**
--- a/woocommerce/inc/ajax-cart.php
+++ b/woocommerce/inc/ajax-cart.php
@@ -4,7 +4,7 @@
    * WooCommerce AJAX cart
    *
    * @package Bootscore
-   * @version 6.4.0
+   * @version 6.5.0
    */
 
 
--- a/woocommerce/js/ajax-cart.js
+++ b/woocommerce/js/ajax-cart.js
@@ -16,7 +16,9 @@ jQuery(function ($) {
   // 1 Handle AJAX Add To Cart
 
   // 1.1 Enable AJAX for single product pages
-  $('form.cart').on('submit', function (e) {
+  // The handler is namespaced so third parties can unbind it in a supported way:
+  // $('form.cart').off('submit.bootscoreAjaxAddToCart');
+  $('form.cart').on('submit.bootscoreAjaxAddToCart', function (e) {
     // Only apply to single product pages, not external products
     if (!$(this).closest('.product-type-external').length) {
       e.preventDefault();
--- a/woocommerce/js/ajax-cart.js
+++ b/woocommerce/js/ajax-cart.js
@@ -1,5 +1,5 @@
 /**
- * AJAX Cart JS - Bootscore v6.4.0
+ * AJAX Cart JS - Bootscore v6.5.0
  *
  * Consists of 4 parts
  * 1. Handle Ajax Add to cart


crftwrk:
@Axos11 if you agree, please merge

Axos11:
@crftwrk I'm looking into it, will take me some days. 

Edit: Not the review of that suggestion but I want to understand the WHY of the general issue :)

